PR #4, opened 2018-12-18
Improve behavior when yielding

carllerche:
Algorithms that depended on yielding could result in loom being unable
to make forwards progress. The workaround implemented in this patch is
not ideal in that it reduces the space searched with regards to atomics,
however atomics need to be revisted anyway.


carllerche:
Pretty sure this made it to master...

PR #14, opened 2019-04-18
Flesh out the std::sync::Arc mock implementation

whentze:
This PR adds the following APIs to `loom::sync::Arc<T>`:

- `Arc::ptr_eq`
- `Arc::strong_count`
- `Default`
- `From<T>`

It also adds some basic docs (the current docs just say "TODO").

Unresolved questions:

- Should we add support for weak refcounting?
- Should we add some tests for this Arc?
- How do we implement the methods that std uses fences for (`try_unwrap` etc.)?

### DIFF
--- a/src/sync/arc.rs
+++ b/src/sync/arc.rs
@@ -29,6 +29,11 @@ impl<T> Arc<T> {
             })
         }
     }
+
+    /// Gets the number of strong (`Arc`) pointers to this value.
+    pub fn strong_count(this: &Self) -> usize {
+        this.inner.ref_cnt.load(SeqCst)
+    }
 }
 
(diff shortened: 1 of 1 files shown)


seanmonstar:
Hm, seems a CI build got lost, there's nothing for me to "restart" it. Could you force push to trigger CI again?

whentze:
I triggered CI again, but it still gets stuck. I believe this is an instance of [travis-ci/#10204]](https://github.com/travis-ci/travis-ci/issues/10204), as #12 is stuck at the same check.

carllerche:
Fun fun... we should migrate to azure pipelines.

carllerche:
I tried restarting travis. If that doesn't work, I will force the merge.

PR #30, opened 2019-08-01
implement atomic int macro and add pub type AtomicU64

### LINKED ISSUE #13: Add support for AtomicBool, AtomicU64 etc.
Everything is in the title :) As far as I can see on docs.rs the `Atomic<>` type is not public, meaning that these cannot be implemented outside of `loom`.

JoshMcguigan:
Partially implementation of #13 

I wanted to try using a macro to reduce code duplication of the usize/u32/u64 atomic types. This MR adds the `AtomicU64` type. Feel free to close if you'd rather not use a macro here.


carllerche:
Woops, I missed the notification for this PR. The impl looks good, but it looks like there is a conflict :(

carllerche:
Thanks for the work :+1:

JoshMcguigan:
Thanks for the follow up. I've rebased this branch against the latest upstream master to resolve the conflict.

Issue #37, opened 2019-08-13
Possible false positive `CausalCell` test case


phlip9:
I've been playing around with loom a bit and wrote this small sanity check seems to pass when it should not:

```rust
#[test]
#[should_panic]
fn causal_cell_race() {
    loom::model(|| {
        let x = Arc::new(CausalCell::new(1_u32));
        let y = Arc::clone(&x);

        let th1 = thread::spawn(move || {
            x.with_mut(|v| unsafe { *v += 1 });
        });

        y.with_mut(|v| unsafe { *v += 10 });

        th1.join().unwrap();

        let v = y.with_mut(|v| unsafe { *v });
        assert_eq!(12, v);
    });
}
```

My intention is for loom's `CausalCell` to catch the two threads entering the same "critical section" by simultaneously acquiring a `*mut u32`; however, the test case seems to pass.

Interestingly, with a small change loom is able to catch the issue:

```rust
#[test]
#[should_panic]
fn causal_cell_race_fixed() {
    loom::model(|| {
        let x = Arc::new(CausalCell::new(1_u32));
        let y = Arc::clone(&x);
        let z = Arc::clone(&x);

        let th1 = thread::spawn(move || {
            x.with_mut(|v| unsafe { *v += 1 });
        });

        let th2 = thread::spawn(move || {
            y.with_mut(|v| unsafe { *v += 10 });
        });

        th1.join().unwrap();
        th2.join().unwrap();

        let v = z.with_mut(|v| unsafe { *v });
        assert_eq!(12, v);
    });
}
```

The output shows an error as expected:

```
thread 'causal_cell_race_fixed' panicked at 'cell=VersionVec { versions: [0, 1, 0, 0] }; thread=VersionVec { versions: [0, 0, 1, 0] }', src/sync/causal/cell.rs:74:13
```

I could also be misunderstanding the loom API (very likely). Anyway, great work with loom--I think this type of tooling is very needed.

carllerche:
Thanks for the repro. Just for my sanity, could you provide the full file (with imports).

phlip9:
For sure. This is just appended verbatim to `loom/tests/causal_cell.rs`, but I'll open a PR.

PR #38, opened 2019-08-13
[WIP] Test false positive causal cell race

phlip9:
see issue #37

### DIFF
--- a/src/rt/execution.rs
+++ b/src/rt/execution.rs
@@ -67,15 +67,17 @@ impl Execution {
     /// Create state to track a new thread
     pub fn new_thread(&mut self) -> thread::Id {
         let thread_id = self.threads.new_thread();
+        let active_id = self.threads.active_id();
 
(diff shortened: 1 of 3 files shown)


carllerche:
Thanks :+1: my best guess here is `thread::spawn` is not bumping the causality version vector. If you want to take a look, see how other points of concurrency bump the version vec and you can probably attempt a fix.

This is the field: https://github.com/carllerche/loom/blob/master/src/rt/thread.rs#L21

carllerche:
Thanks :+1: I pushed a fix to the branch. I would appreciate any further tests you may write :)

phlip9:
Ah great! I just found this issue, but it looks like you fixed it first :)
I'll close the issue as well.

carllerche:
Ah, sorry about that. I ended up doing a context switch to loom last
night to merge a bunch of other PRs too...

PR #40, opened 2019-08-14
Add Condvar::notify_all

seanmonstar:


src/sync/condvar.rs
```diff
@@ -62,4 +62,16 @@ impl Condvar {
             th.unpark();
         }
     }
+
+    /// Wakes up all blocked threads on this condvar.
+    pub fn notify_all(&self) {
+        self.object.branch();
+
+        // Grab a mut lock for sanity of `th.unpark()`...
+        let mut waiters = self.waiters.borrow_mut();
```
carllerche: What is the lock for?

seanmonstar: I'm not as familiar with the whole codebase, so I wasn't sure if unparking a thread here could execute some code that might try to repark it, and thus modify the waiters while iterating. Holding the mutable borrow here at least means it will panic if that were the case.

carllerche: It should be fine. `unpark` only flags the thread as runnable for the next scheduling pass. I believe `branch()` will do the potential context switch.

That said, this isn't a lock, just a ref cell. I think this should be fine.

PR #57, opened 2019-09-20
Add initial atomic::fence support

carllerche:
This only adds support for `atomic::fence(Acquire)` applying to previous loads in the same thread.


carllerche:
Relevant change is here: https://github.com/carllerche/loom/pull/57/files#diff-69c88ce9c3d4efcec2bbd77a0380b1deR113-R137

Everything else is refactoring.

PR #58, opened 2019-09-23
Refactor runtime to use an ECS

carllerche:
An ECS is an entity component system. It is a strategy for storing
objects. Objects are stored in a single vec and "pointers" are an index
into the vector. This allows for complex interactions without making the
borrow checker unhappy.

This refactor also introduces a new primitive: Notify. The notify
primitive simulates the wake / notify pattern. The thread join handle is
updated to use a notify internally and oneshot is removed. Notify can be
initialized to either use SeqCst ordering or acquire / release.


src/rt/mutex.rs
```diff
@@ -0,0 +1,143 @@
...
+pub(crate) struct Mutex {
+    obj: Object,
+}
+
+#[derive(Debug)]
+pub(super) struct State {
+    /// If the mutex should establish sequential consistency.
+    seq_cst: bool,
+
+    /// `true` when the mutex is currently in the locked state.
```
hawkw: Nit: I think this comment is no longer accurate?

src/rt/object.rs
```diff
@@ -53,106 +51,106 @@ pub(crate) enum Action {
...
-    pub(crate) fn new() -> Set {
-        Set { objects: vec![] }
+impl Store {
+    /// Create a new, empty, object store
+    pub(crate) fn new() -> Store {
+        Store { entries: vec![] }
+    }
+
+    /// Insert a new atomic object into the store.
+    pub(super) fn insert_atomic(&mut self, state: atomic::State) -> Object {
```
hawkw: It seems like there's some code here (e.g. `insert_atomic`, `insert_mutex`, insert_condvar`, ...) that could potentially be de-duplicated using traits, maybe
```rust
pub(super) fn insert(&mut self, impl Into<Entry>) -> Object
```
or something?

Not a blocker though; I'm not sure if this actually matters.

src/rt/store.rs
```diff
@@ -0,0 +1,4 @@
+
```
hawkw: Is this file still used? I don't see a `mod store` in `rt/mod.rs`; and it looks like the `Store` type now lives in `rt/object.rs`.

src/rt/object.rs
carllerche: I think a trait strategy could reduce some boilerplate. I opted to punt for this PR.

src/rt/store.rs
carllerche: good catch :+1:

src/rt/mutex.rs
carllerche: Thanks. I updated the comment.

PR #62, opened 2019-09-24
Add mock thread locals

hawkw:
When running a test that uses `loom`'s mock threads, all the mock
threads in the test are actually run on the same OS thread. This means
that code which uses thread local storage will not behave the same
inside of a `loom` test — every `loom` "thread" will actually see the
same version of the thread-local.

This branch adds a mock implementation of the Rust standard library's
`std::thread::LocalKey` type to `loom::thread`, and a mock version of
the `thread_local!` macro that expands to a mock `loom` `LocalKey`.

This is implemented by adding a hash map of `LocalKey`static addresses
to  `Box<dyn Any>`s in the `loom::rt::thread::Thread` type, and having
the mock `LocalKey` type access the current thread from the runtime, and
use its address to index the current thread's map. The boxed `Any` value
is then downcast to the expected type, which should always succeed since
the thread-local is constructed with an initializer returning that type.

Unfortunately, it was necessary to make the mock `LocalKey`'s fields
`#[doc(hidden)] pub`, since we cannot currently write a `const fn`
constructor for it, as `const fn`s with function pointer arguments
aren't currently stable. This can be resolved when function pointer args
to `const fn`s stabilize.

I've also added a test that each `loom` mock thread accesses its own
version of the mock thread local.

Signed-off-by: Eliza Weisman <eliza@buoyant.io>

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -149,3 +149,38 @@ if_futures! {
 pub fn __debug_enabled() -> bool {
     rt::execution(|e| e.log)
 }
+
+/// Mock version of `std::thread_local!`.
+// This is defined *after* all other code in `loom`, since we use
+// `scoped_thread_local!` internally, which uses the `std::thread_local!` macro
+// without namespacing it. Defining this after all other `loom` modules
+// prevents internal code from accidentally using the mock thread local instead
+// of the real one.
+#[macro_export]
+macro_rules! thread_local {
+    // empty (base case for the recursion)
+    () => {};
+
+    // process multiple declarations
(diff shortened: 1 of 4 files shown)


carllerche:
Looks good to me.

One question: are locals dropped when the loom thread completes? The loom thread state stays around for the whole execution even when the thread itself terminated. I _believe_ there is a `rg` function that is called when a thread completes. You could probably force drop locals at that point?

hawkw:
@carllerche I've made some changes to this branch, including running destructors for thread locals when the thread terminates. I've also fixed a bug where loom would panic when `LocalKey::with` is called inside of another call to `with`. If you have the time, it would be great to get another review on the current state of this PR.

PR #69, opened 2019-09-26
switch to a std::thread based scheduler.

carllerche:
The generator based scheduler resulted in undefined behavior and SEGVs.
Unfortunately, this scheduler is much slower.


carllerche:
Also, the scheduler implementation is pretty much hacked together as I didn't have much time to spend on it.

carllerche:
Unfortunately, it looks like this is failing in CI (but not locally).

carllerche:
The bug was in loom itself and has since been fixed.

PR #73, opened 2019-10-03
Check for leaks

carllerche:
A first pass at adding some leak checking to loom + other improvements.

`Arc` is promoted to a full "object". This allows optimizing the behavior of the mocked arc with the loom scheduler. Now, using a mocked `Arc` requires less permutations as loom is able to "see" through it. Also, by promoting it to an object, at the end of each execution, all arcs are checked to see if they were correctly freed. In the event that they were not freed, the execution fails.

The waker behind `block_on` is updated to use a mocked `Arc`. This allows catching leaking wakers.

Mocked versions of `std::alloc::{alloc, dealloc}` are added. These do not impact the scheduler or permutations, but it allows tracking correct usage of allocation & deallocation within the context of an execution. If an allocation leaks, loom fail the execution.

A `loom::alloc::Track<T>` type is added to allow tracking arbitrary values and ensure that the drop handlers are called. This functions in a similar way as allocation tracking.

Finally, `Notify` is improved to permute on spurious wake ups. When waiting on a notify, loom will first try the execution without any spurious wake ups, then it will try the execution with a spurious wake up.

`futures::block_on` is updated to remove the double poll simulating spurious wake ups as doing so could hide bugs.


carllerche:
There is also more general refactoring done.

src/sync/arc.rs
```diff
@@ -1,45 +1,42 @@
...
+                obj: rt::Arc::new(),
             }),
         }
     }
 
     /// Gets the number of strong (`Arc`) pointers to this value.
-    pub fn strong_count(this: &Self) -> usize {
-        this.inner.ref_cnt.load(SeqCst)
+    pub fn strong_count(_this: &Self) -> usize {
+        unimplemented!("no tests checking this? DELETED!")
```
carllerche: 🤔

src/future/mod.rs
```diff
@@ -32,26 +33,42 @@ where
...
-        }
-
-        notify.notify.wait();
+        notify.wait();
     }
 }
 
-struct NotifyWaker {
-    notify: rt::Notify,
+pub(super) fn waker_vtable() -> &'static RawWakerVTable {
```
hawkw: Nit/TIOLI: can't this just be a static rather than a function?

src/rt/alloc.rs
```diff
@@ -0,0 +1,64 @@
...
+
+        let allocation = Allocation { obj };
+
+        let prev = execution.raw_allocations.insert(ptr as usize, allocation);
+        assert!(prev.is_none(), "pointer already tracked");
+    });
+}
+
+/// Track a raw deallocation
+pub(crate) fn dealloc(ptr: *mut u8) {
```
hawkw: Here's a thought (not a blocker for this PR): it would be really cool if `loom`'s leak-detection code could eventually capture the backtrace where an object was allocated from and display it to the user if that allocation was leaked. Could be worth doing in a follow-up...

src/rt/alloc.rs
```diff
@@ -0,0 +1,64 @@
...
+        rt::execution(|execution| {
+            let state = self.obj.alloc(&mut execution.objects);
+            state.is_dropped = true;
+        });
+    }
+}
+
+impl State {
+    pub(super) fn check_for_leaks(&self) {
+        assert!(self.is_dropped, "object leaked");
```
hawkw: Since this is an assert, it will panic immediately when the first leaked allocation is checked. It seems like we might (eventually) want to change this to return a `bool` or something, so that `Execution::check_for_leaks` can count the number of leaked `Arc`s and raw allocations, and panic if either are non-zero, so that all leaks that occurred during an execution can be reported.

This would be especially nice if we also started tracking unique metadata (e.g. stack frames) per allocated object.

src/rt/notify.rs
```diff
@@ -71,10 +79,29 @@ impl Notify {
...
+        let (notified, spurious) = rt::execution(|execution| {
+            let state = self.get_state(&mut execution.objects);
+
+            let spurious = if state.spurious && !state.did_spur {
+                execution.path.branch_spurious()
+            } else {
+                false
+            };
+
+            if spurious {
```
hawkw: Nit/TIOLI: could be simplified to 
```rust
state.did_spur = state.did_spur || spurious;
```

src/rt/object.rs
```diff
@@ -32,27 +34,43 @@ enum Entry {
...
     Opaque,
 }
 
 impl Object {
+    pub(super) fn alloc(self, store: &mut Store) -> &mut alloc::State {
+        assert_eq!(self.execution_id, store.execution_id);
+
+        match &mut store.entries[self.index] {
+            Entry::Alloc(v) => v,
+            _ => panic!(),
```
hawkw: Nit: should this be `unreachable!()`?

src/sync/arc.rs
```diff
@@ -78,7 +75,13 @@ impl<T> Clone for Arc<T> {
 
 impl<T> Drop for Arc<T> {
     fn drop(&mut self) {
-        self.inner.ref_cnt.fetch_sub(1, AcqRel);
+        if self.inner.obj.ref_dec() {
+            assert_eq!(
+                1,
+                std::sync::Arc::strong_count(&self.inner),
+                "something odd is going on"
```
hawkw: 🙃

PR #74, opened 2019-10-09
Misc fixes

carllerche:
Collecting up a bunch of fixes.

This includes a fix to the thread local implementation. All "hooks" must be called from outside of the `rt::execution` closure.

### DIFF
--- a/src/rt/thread.rs
+++ b/src/rt/thread.rs
@@ -27,7 +27,7 @@ pub(crate) struct Thread {
     /// Number of times the thread yielded
     pub yield_count: usize,
 
-    locals: HashMap<LocalKeyId, LocalValue>,
+    locals: LocalMap,
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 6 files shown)


src/rt/thread.rs
```diff
@@ -126,10 +128,17 @@ impl Thread {
 
     pub(crate) fn set_terminated(&mut self) {
         self.state = State::Terminated;
+    }
+
+    pub(crate) fn drop_locals(&mut self) -> Box<dyn std::any::Any> {
```
hawkw: Any reason this doesn't just return a `Vec<LocalValue>`, or is this just to hide the actual type from calling code so that all it can do is drop it?

hawkw: Also: might be good to have a comment here explaining why this doesn't actually drop anything :)

carllerche: It was a private type, so I opted to hide it... I didn't think deeply about it.

PR #88, opened 2019-10-29
Add support for RwLock

### LINKED ISSUE #21: Add support for RwLock
I notice that `loom::sync` has a [mock implementation of `std::sync::Mutex`](https://github.com/carllerche/loom/blob/master/src/sync/mutex.rs) but lacks a mock `std::sync::RwLock`. It would be nice to be able to test code using read-write locks with `loom`.

pop:
Fixes #21 

Oh snap when did this move to the Tokio org? Well, all the same: I added support for RwLock!

I don't know if this is done, but the tests (which I got from the RwLock docs) pass so I'm happy.

I would love some feedback. The `is_read_locked()` and `is_write_locked()` functions cause "Already Locked" panics so I am not using them at the moment. Let me know if you want more info on that.

cc: @pickfire ?


src/rt/rwlock.rs
```diff
@@ -0,0 +1,269 @@
...
+            RwLock { obj }
+        })
+    }
+
+    fn get_state<'a>(&self, objects: &'a mut object::Store) -> &'a mut State {
+        self.obj.rwlock_mut(objects).unwrap()
+    }
+}
+
+impl RwLock {
```
pickfire: I wonder why is this separated from the `RwLock` impl above?

pickfire:
@pop Nice, I believe I may be lack of experience to review this but probably @hawkw could help.

.gitignore
```diff
@@ -1,2 +1,6 @@
 target
 Cargo.lock
+
+# Vim
+.*.swp
+.*.swo
```
hawkw: Nit: ideally, we'd add this in a separate PR so it's not part of the diff for adding the mock `RwLock`.

src/rt/rwlock.rs
```diff
@@ -0,0 +1,269 @@
...
+use std::collections::HashSet;
+use std::sync::atomic::Ordering::{Acquire, Release};
+
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub(crate) struct RwLock {
+    obj: Object,
+}
+
+#[derive(Debug, PartialEq)]
+enum RwLockType {
```
hawkw: Nit: i'm not a huge fan of this naming...but I'm not really sure what would be better off the top of my head. Maybe `Locked`?

src/rt/rwlock.rs
```diff
@@ -0,0 +1,269 @@
...
+    Write(thread::Id),
+}
+
+#[derive(Debug)]
+pub(super) struct State {
+    /// If the rwlock should establish sequential consistency.
+    seq_cst: bool,
+
+    /// `Some` when the rwlock is in the locked state.
+    /// The `thread::Id` references the thread that currently holds the rwlock.
```
hawkw: I think this comment could be made clearer, since the `lock` value could either be a single thread ID holding the write lock, or multiple thread IDs holding read locks.

src/rt/rwlock.rs
```diff
@@ -0,0 +1,269 @@
...
+                lock: None,
+                last_access: None,
+                synchronize: Synchronize::new(execution.max_threads),
+            });
+
+            RwLock { obj }
+        })
+    }
+
+    fn get_state<'a>(&self, objects: &'a mut object::Store) -> &'a mut State {
```
hawkw: Nit/style: I'd prefer it if we organized this impl block with all the `pub(crate)` methods first, followed by the private methods...not going to block merging the PR on that, though.

src/rt/rwlock.rs
```diff
@@ -0,0 +1,269 @@
...
+        super::execution(
+            |execution| match self.get_state(&mut execution.objects).lock {
+                Some(RwLockType::Read(_)) => true,
+                _ => false,
+            },
+        )
+    }
+}
+
+impl RwLock {
```
hawkw: Why are there multiple `impl` blocks here? Can these be combined into one impl block?

src/rt/rwlock.rs
```diff
@@ -0,0 +1,269 @@
...
+                None => {
+                    let mut threads: HashSet<thread::Id> = HashSet::new();
+                    threads.insert(thread_id);
+                    threads
+                }
+                Some(RwLockType::Read(current)) => {
+                    let mut threads: HashSet<thread::Id> = HashSet::new();
+                    threads.extend(current);
+                    threads.insert(thread_id);
+                    threads
```
hawkw: Is there a reason we create a new `HashSet` here, rather than something like this:
```rust
let thread_set = match state.lock {
    // ...
    Some(RwLockType::Read(ref mut current)) => {
        current.insert(thread_id);
        current
    },
    // ...
}

File: src/rt/rwlock.rs [hunk 8]
```diff
@@ -0,0 +1,269 @@
...
+            };
+
+            dbg!(state.synchronize.sync_load(&mut execution.threads, Acquire));
+
+            if state.seq_cst {
+                // Establish sequential consistency between locks
+                execution.threads.seq_cst();
+            }
+
+            // Block all other threads attempting to acquire rwlock
```
hawkw: Nit/TIOLI: seems like this code is repeated enough that it could maybe be factored out into its own function...not a blocker.

src/rt/rwlock.rs
```diff
@@ -0,0 +1,269 @@
...
+
+            true
+        })
+    }
+
+    fn is_write_locked(&self) -> bool {
+        super::execution(
+            |execution| match self.get_state(&mut execution.objects).lock {
+                Some(RwLockType::Write(_)) => true,
+                Some(RwLockType::Read(_)) | None => false,
```
hawkw: TIOLI:
```suggestion
                _ => false,
```

src/rt/rwlock.rs
```diff
@@ -0,0 +1,269 @@
...
+    /// Tracks write access to the rwlock.
+    last_access: Option<Access>,
+
+    /// Causality transfers between threads
+    synchronize: Synchronize,
+}
+
+impl RwLock {
+    /// Common RwLock function
+    pub(crate) fn new(seq_cst: bool) -> RwLock {
```
hawkw: As far as I can tell, the `rt::RwLock` type is only ever constructed in `sync::RwLock::new`, and it always passes `true` here. Will we ever construct a `rt::RwLock` with `seq_cst=false`? If not, should we just remove the bool and always establish sequential consistency?

src/sync/rwlock.rs
```diff
@@ -0,0 +1,140 @@
...
+pub struct RwLock<T> {
+    object: rt::RwLock,
+    data: std::sync::RwLock<T>,
+}
+
+/// Mock implementation of `std::sync::RwLockReadGuard`
+#[derive(Debug)]
+pub struct RwLockReadGuard<'a, T> {
+    lock: &'a RwLock<T>,
+    data: Option<std::sync::RwLockReadGuard<'a, T>>,
```
hawkw: It's not immediately obvious why this is an `Option`. I think it's it so that the `Drop` impl can drop the `std` guard and release the `std` lock _before_ releasing the `loom` mock lock, because that might cause another thread to acquire the lock? 

I think a comment explaining this (either here on the field or in the `Drop` impl) would be very helpful.

src/rt/rwlock.rs
pop: It was just for readability. One block is the reading methods, one block is the writing methods, third for the rest. I can combine them 😄

src/rt/rwlock.rs
pop: I didn't go down that route because it quickly unraveled into needing to specify lifetime parameters everywhere and I am not adept enough at Rust to do that... with grace. I'll dig into it though and see if I can make it work.

src/rt/rwlock.rs
pickfire: Shouldn't it be called `state` according to API guidelines?

src/rt/rwlock.rs
pickfire: Oh, I was also asking about that in another commit.

src/rt/rwlock.rs
pop: An identical method is called `get_state` in Mutex FWIW.

pop:
I'm taking a second look at some of what I implemented. In trying to factor out some code I think I found a bug in my implementation. I'll remove the WIP title when this is ready for review again.

src/rt/rwlock.rs
hawkw: Hmm, if it isn't easily changed, I don't think this is a blocker, it just seems kind of inefficient...

pop:
This is ready for a final review... or at least I think it is right now... so throw your feedback at me (again)!

Some feedback I haven't taken:
* A part of the code looks like I should be able to edit a HashMap but instead I create a new one. I need a bit more hand-holding to get that refactor done. I don't know how to make rustc happy. When I end up putting lifetime specifiers in stuff that is three steps removed from the change I'm making I start to think "this is not a good refactor".
* There's some repeated code. I refactored out that loop and a similar loop for symmetry. I'm not super sure about that choice, but the loops had a bug so I figured putting them close to eachother in the code might make it easier to contrast the two.

hawkw:
@pop re:
> A part of the code looks like I should be able to edit a HashMap but instead I create a new one. I need a bit more hand-holding to get that refactor done. I don't know how to make rustc happy. 

what do you think about something like this:
```rust
            // Set the lock to the current thread
            let mut already_locked = false;
            state.lock = match state.lock.take() {
                None => {
                    let mut threads: HashSet<thread::Id> = HashSet::new();
                    threads.insert(thread_id);
                    Some(Locked::Read(threads))
                }
                Some(Locked::Read(threads)) => {

                    threads.insert(thread_id);
                    Some(Locked::Read(threads))
                }
                Some(Locked::Write(writer)) => {
                    already_locked = true;
                    Some(Locked::Write(writer))
                }
            };

            if already_locked {
                return false;
            }
```

src/rt/rwlock.rs
```diff
@@ -0,0 +1,243 @@
...
+enum Locked {
+    Read(HashSet<thread::Id>),
+    Write(thread::Id),
+}
+
+#[derive(Debug)]
+pub(super) struct State {
+    /// A single `thread::Id` when Write locked.
+    /// A set of `thread::Id` when Read locked.
+    lock: Option<Locked>,
```
pickfire: Why does this have to be `Option<Locked>`? Could it be `Locked` and initialized at first use?

pop: If no threads are reading or writing the RwLock then it doesn't make sense for it to be in either `Locked::Read` or `Locked::Write`, so I think `None` makes the most sense.

pop:
Thanks @hawkw and @pickfire for the review(s)! I'm learning a lot :smile:

src/rt/rwlock.rs
pickfire: Ah, so similar to the following
```rust
#[derive(Debug, PartialEq)]
enum Locked {
    Read(HashSet<thread::Id>),
    Write(thread::Id),
    None,
}
```
I guess `Option<Locked>` makes more sense now but I wonder if the size is any different from `Option<Locked>` compared to `Locked`.

src/rt/object.rs
```diff
@@ -106,6 +107,15 @@ impl Object {
             _ => None,
         }
     }
+
+    pub(super) fn rwlock_mut(self, store: &mut Store) -> Option<&mut rwlock::State> {
+        assert_eq!(self.execution_id, store.execution_id);
+
+        match &mut store.entries[self.index] {
```
pickfire: Why is the `&mut` required here? I rarely sees a `&mut` for match statement, I wonder what it does.

tests/mutex.rs
```diff
@@ -1,4 +1,4 @@
-#![deny(warnings, rust_2018_idioms)]
+// #![deny(warnings, rust_2018_idioms)]
```
pickfire: Why is this change necessary?

src/rt/rwlock.rs
pop: Hmm. I don't know about that. I was more focused on the expressiveness of the implementation and not the size impact. I can change it to your suggestion if we're concerned about it. Your suggestion seems pretty expressive too -- "the RwLock can be in one of three states..." makes sense too.

src/rt/object.rs
pop: In this case I was following the other `*_mut` methods. `mutext` does the same thing so I copy-pasted it. I can look into it, but I think consistency (all of the `*_mut` methods look _exactly_ the same) is important here so we should update all methods -- or I'll need to figure out why `rwlock` is different from `mut`.

pop: * To answer your question I think we `&mut` here because we are returning an `Option<&mut rwlock::State>` in the `match`. If I try to get rid of the `&mut` here I get messages like this:
```rs
error[E0308]: mismatched types
   --> src/rt/object.rs:115:38
    |
115 |             Entry::RwLock(v) => Some(v),
    |                                      ^
    |                                      |
    |                                      expected mutable reference, found struct `rt::rwlock::State`
    |                                      help: consider mutably borrowing here: `&mut v`
    |
    = note: expected type `&mut rt::rwlock::State`
               found type `rt::rwlock::State`
```

but I can't return `Some(&mut v)` because we're not borrowing as mutable.

tests/mutex.rs
pop: It wasn't 😅 -- just a typo.

src/rt/rwlock.rs
hawkw: @pickfire I also thought about mentioning that, but I saw that `Mutex` had a function called `get_state`, and I thought consistency was good... 🤷‍♀

src/rt/rwlock.rs
pickfire: No, I am not sure whether it affects the size, just thought about it, we need to verify it with probably something like `std::mem::size_of<Type>()`.

src/rt/rwlock.rs
```diff
@@ -182,22 +182,28 @@ impl RwLock {
...
-                    Some(Locked::Read(new))
+                Some(Locked::Read(mut threads)) => {
+                    threads.insert(thread_id);
+                    Some(Locked::Read(threads))
+                }
+                Some(Locked::Write(writer)) => {
+                    already_locked = true;
+                    Some(Locked::Write(writer))
                 }
-                Some(Locked::Write(_)) => return false,
```
pickfire: Why not just keep
```rust
                 Some(Locked::Write(_)) => return false,
```
and ignore
```rust

                 Some(Locked::Write(writer)) => {
                    already_locked = true;
```

hawkw: @pickfire because of the [call to `Option::take` on line 186](https://github.com/tokio-rs/loom/pull/88/commits/cd84f9dd31438c588f53ef507e50ac3b1f952efc#diff-e9106da480ca203b688aa13f79b66051R186) that would be incorrect. Since we've `take`n the previous value, `state.lock` is currently `None`. If we return false here, we leave the state as `None`, incorrectly releasing the lock.

pop:
I think we're just waiting on a review from @carllerche, is that right @hawkw @pickfire?

hawkw:
> I think we're just waiting on a review from @carllerche, is that right @hawkw @pickfire?
@pop as far as I know, that's right! This looks good to me!

pickfire:
Yup, looks good to me too.

pop:
Rebased against master so there shouldn't be any conflicts. Thanks @hawkw (via #105)!

carllerche:
I merged master and fixed the conflicts :+1:

carllerche:
Sorry for the delay, thanks for doing the work!

carllerche:
Btw, there is a loom channel on the Tokio discord (https://discord.gg/tokio). If something is blocked on me, feel free to ping me there.

PR #92, opened 2019-10-29
Remove the redundant `init_threads` field

nkaretnikov:


nkaretnikov:
I haven't checked when this was introduced and why, but this field appears to be redundant. Or would you still like to use it in the future for something?

nkaretnikov:
`git blame` shows this was introduced by `b9d5a95cb102d849257d2c28bf8a6c24fce943e8`. I guess it was added while prototyping and just wasn't removed.

adrianheine:
This was finally removed in 88e4225b4e9e88d0f050cd7061d8e7bc02d16d35.

PR #94, opened 2019-11-03
Some performance improvements

ztlpn:
A couple of small changes to reduce memory allocations:
* switch `last_dependent_accesses` to internal iteration to avoid boxing iterators
* reuse VersionVecs of last accesses

As a benchmark, with these changes running time of tokio-executor loom tests went down from 21.0s (with loom 0.2.12) to 16.6s on my machine.

### DIFF
--- a/src/rt/access.rs
+++ b/src/rt/access.rs
@@ -19,6 +19,14 @@ impl Access {
         self.dpor_vv.set(version);
     }
 
+    pub(crate) fn set_or_create(access: &mut Option<Self>, path_id: usize, version: &VersionVec) {
+        if let Some(access) = access.as_mut() {
+            access.set(path_id, version);
+        } else {
+            *access = Some(Access::new(path_id, version));
+        }
+    }
+
     /// Location in the path
     pub(crate) fn path_id(&self) -> usize {
         self.path_id
--- a/src/rt/access.rs
+++ b/src/rt/access.rs
@@ -14,6 +14,11 @@ impl Access {
         }
     }
 
+    pub(crate) fn set(&mut self, path_id: usize, version: &VersionVec) {
+        self.path_id = path_id;
+        self.dpor_vv.set(version);
+    }
+
     /// Location in the path
     pub(crate) fn path_id(&self) -> usize {
         self.path_id
--- a/src/rt/arc.rs
+++ b/src/rt/arc.rs
@@ -134,17 +134,9 @@ impl State {
     }
 
     pub(super) fn set_last_access(&mut self, action: Action, path_id: usize, version: &VersionVec) {
-        let set_or_create = |access: &mut Option<Access>| {
-            if let Some(access) = access.as_mut() {
-                access.set(path_id, version);
-            } else {
-                *access = Some(Access::new(path_id, version));
-            }
-        };
-
         match action {
-            Action::RefInc => set_or_create(&mut self.last_ref_inc),
-            Action::RefDec => set_or_create(&mut self.last_ref_dec),
+            Action::RefInc => Access::set_or_create(&mut self.last_ref_inc, path_id, version),
+            Action::RefDec => Access::set_or_create(&mut self.last_ref_dec, path_id, version),
         }
     }
 }
--- a/src/rt/arc.rs
+++ b/src/rt/arc.rs
@@ -1,6 +1,6 @@
 #![allow(warnings)]
 use crate::rt::object::Object;
-use crate::rt::{self, Access, Synchronize};
+use crate::rt::{self, Access, Synchronize, VersionVec};
 
 use std::sync::atomic::Ordering::{Acquire, Release};
 
@@ -124,7 +124,7 @@ impl State {
     pub(super) fn for_each_last_dependent_access(
         &self,
         action: Action,
-        mut f: impl FnMut(&Access)
+        mut f: impl FnMut(&Access),
     ) {
         match action {
             // RefIncs are not dependent w/ RefDec, only inspections
@@ -133,10 +133,18 @@ impl State {
         }
     }
 
-    pub(super) fn set_last_access(&mut self, action: Action, access: Access) {
+    pub(super) fn set_last_access(&mut self, action: Action, path_id: usize, version: &VersionVec) {
+        let set_or_create = |access: &mut Option<Access>| {
+            if let Some(access) = access.as_mut() {
+                access.set(path_id, version);
+            } else {
+                *access = Some(Access::new(path_id, version));
+            }
+        };
+
         match action {
-            Action::RefInc => self.last_ref_inc = Some(access),
-            Action::RefDec => self.last_ref_dec = Some(access),
+            Action::RefInc => set_or_create(&mut self.last_ref_inc),
+            Action::RefDec => set_or_create(&mut self.last_ref_dec),
         }
     }
 }
--- a/src/rt/arc.rs
+++ b/src/rt/arc.rs
@@ -121,14 +121,15 @@ impl State {
         assert_eq!(0, self.ref_cnt, "Arc leaked");
     }
 
-    pub(super) fn last_dependent_accesses<'a>(
-        &'a self,
+    pub(super) fn for_each_last_dependent_access(
+        &self,
         action: Action,
-    ) -> Box<dyn Iterator<Item = &'a Access> + 'a> {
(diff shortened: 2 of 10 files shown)


src/rt/atomic.rs
```diff
@@ -144,24 +144,30 @@ pub(crate) fn fence(order: Ordering) {
...
+        _action: Action,
+        mut f: impl FnMut(&Access),
+    ) {
+        self.last_load.iter().for_each(&mut f);
+        self.last_store.iter().for_each(&mut f);
     }
 
-    pub(super) fn set_last_access(&mut self, action: Action, access: Access) {
+    pub(super) fn set_last_access(&mut self, action: Action, path_id: usize, version: &VersionVec) {
+        let set_or_create = |access: &mut Option<Access>| {
```
hawkw: Nit, take it or leave it: would be nice if this repeated code was factored out, maybe a method on `&mut Option<Access>` or something? Not a blocker.

hawkw: If `Access` implemented `clone_from`, it looks like `Option::clone_from` would do something close to what we want out of the box: https://doc.rust-lang.org/src/core/option.rs.html#1191-1196

src/rt/condvar.rs
```diff
@@ -86,11 +86,15 @@ impl Condvar {
 }
 
 impl State {
-    pub(super) fn last_dependent_accesses<'a>(&'a self) -> Box<dyn Iterator<Item = &Access> + 'a> {
-        Box::new(self.last_access.iter())
+    pub(super) fn for_each_last_dependent_access(&self, f: impl FnMut(&Access)) {
```
hawkw: It looks to me that rather than changing this to a `for_each`, this method could just return an `impl Iterator` rather than a `Box`?

src/rt/atomic.rs
```diff
@@ -144,24 +144,30 @@ pub(crate) fn fence(order: Ordering) {
...
-    pub(super) fn last_dependent_accesses<'a>(
-        &'a self,
-        action: Action,
-    ) -> Box<dyn Iterator<Item = &'a Access> + 'a> {
-        match action {
-            Action::Load => Box::new({ self.last_load.iter().chain(self.last_store.iter()) }),
-            Action::Store => Box::new({ self.last_load.iter().chain(self.last_store.iter()) }),
-            Action::Rmw => Box::new({ self.last_load.iter().chain(self.last_store.iter()) }),
-        }
+    pub(super) fn for_each_last_dependent_access(
```
hawkw: I think that rather than changing this method to a `for_each`, it could just return an `impl Iterator` instead of a box (since all the match arms return an iterator of the same type) --- this would have the same perf benefit of not allocating, but the API would be more flexible and idiomatic.

src/rt/arc.rs
```diff
@@ -121,21 +121,30 @@ impl State {
...
+        mut f: impl FnMut(&Access),
+    ) {
         match action {
             // RefIncs are not dependent w/ RefDec, only inspections
-            Action::RefInc => Box::new([].into_iter()),
-            Action::RefDec => Box::new(self.last_ref_dec.iter()),
+            Action::RefInc => {}
+            Action::RefDec => self.last_ref_dec.iter().for_each(f),
         }
     }
```
hawkw: As an alternative, what do you think about something like this:
```rust
   pub(super) fn last_dependent_accesses<'a>(
        &'a self,
        action: Action,
    ) -> impl Iterator<Item = &'a Access> + 'a> {
        let iter = match action {
            Action::RefInc => None,
            Action::RefDec => Some(self.last_ref_dec.iter()),
        };
        iter.into_iter().flat_map(|i| i)
    }
```
Because the iterator constructed by `Option::flat_map` has the same type regardless of whether it's empty or not, we can return an `impl Iterator`, so we still have the performance benefit of not allocating, but we can continue returning an iterator. This lets downstream code consume it in more complex ways than `for_each`ing, which could be useful in the future.

This is just a thought, feel free to take it or leave it

src/rt/mutex.rs
```diff
@@ -133,11 +133,15 @@ impl Mutex {
 }
 
 impl State {
-    pub(crate) fn last_dependent_accesses<'a>(&'a self) -> Box<dyn Iterator<Item = &Access> + 'a> {
-        Box::new(self.last_access.iter())
+    pub(crate) fn for_each_last_dependent_access(&self, f: impl FnMut(&Access)) {
```
hawkw: Again, I think the box could've just been changed to `impl Iterator`

src/rt/notify.rs
```diff
@@ -130,11 +130,15 @@ impl Notify {
 }
 
 impl State {
-    pub(crate) fn last_dependent_accesses<'a>(&'a self) -> Box<dyn Iterator<Item = &Access> + 'a> {
-        Box::new(self.last_access.iter())
+    pub(crate) fn for_each_last_dependent_access(&self, f: impl FnMut(&Access)) {
```
hawkw: ...you know the drill :)

src/rt/access.rs
```diff
@@ -14,6 +14,11 @@ impl Access {
         }
     }
 
+    pub(crate) fn set(&mut self, path_id: usize, version: &VersionVec) {
```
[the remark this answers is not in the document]
ztlpn: I think it is not possible to do directly without sacrificing much of the benefit - to use `clone_from` we must have an existing instance of `Access` and creating it at the call site of `set_last_access` incurs an additional unnecessary allocation.

One possible way to implement it is to store and update `Access` directly in the `Thread` struct but it feels unnatural (and storing `path_id` can be error-prone as it can get out of sync with the actual path position).

src/rt/atomic.rs
ztlpn: Methods for concrete objects can be made to return `Impl Iterator` (and although chaining does incur some cost, it is very slight). But the clients don't use them - they use the method of `Store` and it still needs to be `for_each_...` so that's not much of an improvement in my opinion. Also, as we have here a kind of polymorphism by hand, I think it makes sense for all methods to have the same structure. I'll defer to your judgment here - if you think that the `Impl Iterator` version is better, I'll implement it.

ztlpn:
@hawkw Thanks for the review and suggestions (didn't know about `clone_from`!) Although I guess they come with their own tradeoffs so we'll have to choose :)

src/rt/arc.rs
```diff
@@ -121,21 +121,22 @@ impl State {
         assert_eq!(0, self.ref_cnt, "Arc leaked");
     }
 
-    pub(super) fn last_dependent_accesses<'a>(
-        &'a self,
+    pub(super) fn for_each_last_dependent_access(
```
carllerche: I **believe** that returning an iterator is no longer necessary. This can **probably** be changed to:

```rust
fn last_dependent_access(&self, action: Action) -> Option<&Access> { ... }
```

src/rt/access.rs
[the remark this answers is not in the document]
hawkw: Ah, yeah, you're right; the `clone_from` behavior is not identical. Nevermind!

src/rt/atomic.rs
hawkw: > But the clients don't use them - they use the method of `Store` and it still needs to be `for_each_...` so that's not much of an improvement in my opinion. 

Ah, yeah — I overlooked that when I left this comment. Since the `Store` method cannot return an iterator, I think this is fine. A comment on these methods might be helpful, though?

src/rt/arc.rs
ztlpn: Yes, every object state except Atomic doesn't need an iterator. Atomic still needs to iterate over `last_load` and `last_store` though. But I like the idea of simply returning an Option for every other object - the code is simpler and it is even a bit faster that way :)

carllerche: ah, i'm actually pretty sure that can be removed... Instead of separate load / store access values, there can be a single one for the entire atomic (the reason there are multiple is a failed attempt at an optimization).

carllerche: Do you want to open a separate PR to do ^^ then we can update this one.

ztlpn: https://github.com/tokio-rs/loom/pull/98

ztlpn:
Combined effect (with #98) on tokio-executor loom tests running time: 20.5s -> 15.2s

PR #100, opened 2019-11-16
Reducing allocations for the Path struct

ztlpn:
Conceptually Path is a stack that gets pushed and popped a lot of times during the test run. Elements of this stack contain small Vecs and VecDeques which are allocated/deallocated a lot of times. We can reduce the number of these allocations by storing thread states and writes in a "SliceVec" - a container that can store slices of different lengths in a single chunk of memory with random access, pushing and popping. After these SliceVecs quickly grow to appropriate size there are no more allocations.

### Performance impact

Using tokio-executor loom tests at revision https://github.com/tokio-rs/tokio/commit/03a9378297c73c2e56a6d6b55db22b92427b850a as a benchmark, execution time compared to loom master fell 15.3s -> 13.6s (11%).

### DIFF
--- a/src/rt/path.rs
+++ b/src/rt/path.rs
@@ -31,7 +31,9 @@ pub(crate) struct Path {
     thread_states: SliceVec<Thread>,
 
     /// Atomic writes
-    writes: Vec<VecDeque<usize>>,
+    writes: SliceVec<usize>,
+    /// i-th element in `current_writes` tracks current position in the i-th slice in `writes`
+    current_writes: Vec<usize>,
 
     /// Tracks spurious notifications
     spurious: Vec<VecDeque<bool>>,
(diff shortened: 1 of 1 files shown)


carllerche:
Similar comment to the bump allocator PR.

I've been working on refactoring internals to use the object store more instead of adhoc vec storage everywhere. If the refactor pans out, it may end up producing similar results to here.

https://github.com/tokio-rs/loom/tree/exp

Once tests pass, I will provide measurements compared to this.

adrianheine:
> Similar comment to the bump allocator PR.
> 
> I've been working on refactoring internals to use the object store more instead of adhoc vec storage everywhere. If the refactor pans out, it may end up producing similar results to here.
> 
> https://github.com/tokio-rs/loom/tree/exp
> 
> Once tests pass, I will provide measurements compared to this.

The mentioned refactoring is 88e4225b4e9e88d0f050cd7061d8e7bc02d16d35 and was merged.

PR #110, opened 2020-03-03
basic backtrace tracking

carllerche:
An initial pass at tracking backtraces during execution. Currently, only
`Arc::new` is tracked, but now that infrastructure is added, tracking
additional events should be easier.

Capturing a backtrace is a an expensive event. The failure case should
be isolated before capturing backtraces.

### DIFF
--- a/src/model.rs
+++ b/src/model.rs
@@ -45,6 +45,15 @@ pub struct Builder {
     /// Defaults to `LOOM_CHECKPOINT_INTERVAL` environment variable.
     pub checkpoint_interval: usize,
 
+    /// When `true`, backtraces are captured on each loom operation.
+    ///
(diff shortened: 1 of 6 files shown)


src/model.rs
```diff
@@ -75,6 +84,8 @@ impl Builder {
             })
             .unwrap_or(DEFAULT_MAX_BRANCHES);
 
+        let backtrace = env::var("LOOM_BACKTRACE").is_ok();
```
hawkw: It seems like if the env var is set, but the feature flag isn't, we should probably print something (or panic) to let the user know that the env var won't do anything?

carllerche: it panics during runtime :shrug:

carllerche: It's a bit of a hack

PR #114, opened 2020-03-24
Add location tracking to Atomic types

seanmonstar:


src/rt/location.rs
```diff
@@ -103,6 +103,9 @@ impl PanicBuilder {
         if let Some(width) = width {
             msg = format!("\n{}", msg);
             for (key, thread, location) in &self.locations {
+                if !location.is_captured() {
```
seanmonstar: Note: I saw a subtraction overflow sometimes here.

src/sync/atomic/atomic.rs
```diff
@@ -18,43 +18,51 @@ where
         Atomic { state }
     }
 
+    #[cfg_attr(loom_nightly, track_caller)]
```
carllerche: I don't think you want to track the caller here. Instead, you probably want to update these fns to take `Location` and track the caller on the wrappers (`AtomicPtr`, `AtomicBool`, `Atomic*`, ..).

hawkw: Yeah, I think that the caller here is always going to be inside loom?

seanmonstar: I don't believe it matters either way, the location is does get correctly propagated to the user's call on `AtomicPtr::whatever()` (visually tested).

PR #118, opened 2020-04-08
Add support for mpsc channels

vakaras:
Adds Loom mocks for `std::sync::mpsc`. Also, adds unimplemented stubs for `std::sync::barrier` that the code that depends on them could be at least compiled.

**Note:** This PR includes the changes from #88, so you probably want to merge it first. **cc**: @pop

### DIFF
--- a/src/sync/mpsc.rs
+++ b/src/sync/mpsc.rs
@@ -1,0 +1,67 @@
+//! A stub for `std::sync::mpsc`.
+
+use crate::rt;
+
+/// Mock implementation of `std::sync::mpsc::channel`.
+pub fn channel<T>() -> (Sender<T>, Receiver<T>) {
+    let (sender_channel, receiver_channel) = std::sync::mpsc::channel();
+    let channel = std::sync::Arc::new(rt::Channel::new());
+    let sender = Sender {
+        object: std::sync::Arc::clone(&channel),
+        sender: sender_channel,
+    };
+    let receiver = Receiver {
+        object: std::sync::Arc::clone(&channel),
+        receiver: receiver_channel,
+    };
+    (sender, receiver)
+}
+
+#[derive(Debug)]
+/// Mock implementation of `std::sync::mpsc::Sender`.
+pub struct Sender<T> {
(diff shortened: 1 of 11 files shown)


carllerche:
I merged #88, would you be able to rebase on that? Thanks.

vakaras:
> I merged #88, would you be able to rebase on that? Thanks.

Thank you for merging #88! I have rebased this PR.

vakaras:
> Looks good, I skimmed and it matches my expectation. I did have one question, which I added inline.

For some reason, I do not see your question. (I also have not received any notification about it.)

carllerche:
@vakaras very strange...

You have one `Synchronize` for the entire channel. I believe synchronization should happen on a message basis, no? So, you would have one `Synchronize` value per message. When you recv a message, you apply the synchronization received for that message.

vakaras:
Good point. Since `mpsc` supports multiple producers, we need to synchronize on a message basis. Thanks for pointing this out! I will update the PR.

vakaras:
Since the [documentation](https://doc.rust-lang.org/std/sync/mpsc/fn.channel.html) claims that `mpsc` channels preserve the order of the messages, I changed the code to have one synchronization point for senders and a synchronization point per message for receivers. As a result, if thread `t1` managed to send `m1` before `t2` sent `m2`, the thread that received `m2`will know that `m1` was already sent and received. However, the receiver of `m1` will know nothing about `m2`.

src/sync/mpsc.rs
```diff
@@ -0,0 +1,67 @@
...
+        }
+    }
+}
+
+#[derive(Debug)]
+/// Mock implementation of `std::sync::mpsc::Receiver`.
+pub struct Receiver<T> {
+    object: std::sync::Arc<rt::Channel>,
+    receiver: std::sync::mpsc::Receiver<T>,
+}
```
carllerche, line 50: This probably should have a `Drop` impl that drains the remaining messages (std has this). Otherwise, I believe the `check_leak` assert will be hit.

vakaras: Added.

PR #119, opened 2020-04-08
fix `UnsafeCell` bug

carllerche:
The thread version vector was not bumped on `UnsafeCell` access. This
caused synchronizations that happened immediately **before** the access
to prevent loom from detecting invalid `UnsafeCell` access.

### DIFF
--- a/src/rt/cell.rs
+++ b/src/rt/cell.rs
@@ -65,7 +65,7 @@ impl Cell {
         }
 
         // Enter the read closure
-        let _reset = rt::execution(|execution| {
+        let _reset = rt::synchronize(|execution| {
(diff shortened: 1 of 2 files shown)


carllerche:
@hawkw like what?

jonhoo:
There's this cool library called [`loom`](https://github.com/tokio-rs/loom/) that we could maybe run the test suite through?

hawkw:
> @hawkw like what?

I'm not sure. If there are operations that we want to ensure always synchronize, we could make them require a token or something that ensures they are part of a synchronize closure? But, I'm not sure how much of the model we can encode in this way, so it may not be that valuable.

PR #122, opened 2020-04-09
Look up atomic locations in the right place

jonhoo:


src/rt/atomic.rs
```diff
@@ -535,7 +531,7 @@ impl State {
         if let Some(loaded) = current.ahead(&self.unsync_loaded_at) {
             location::panic("Causality violation: Concurrent `unsync_load` and mut accesses.")
                 .location("created", self.created_location)
-                .thread("mut", loaded, self.unsync_loaded_locations[loaded])
+                .thread("with_mut", loaded, self.unsync_mut_locations[loaded])
```
carllerche, line 534: I think this tests concurrent `unsync_load`, so the two conflicts should be "unsync load 1" and "unsync load 2"?

jonhoo: Hmm, why would two `unsync_load`s race? I wonder if it's the `ahead` check that's wrong, and should be checking `unsync_mut_at`?

jonhoo:
As far as I can tell, it basically isn't possible to write a test for dc7eb04ce63791475de3167169f8ad6aa71a1c64 without some really unsafe code, since `with_mut` requires `&mut self`, so you wouldn't be able to call `unsync_load` at the same time without some raw pointer magic. I could write that though I suppose.

jonhoo:
Do we _also_ want a test for concurrent unsynced loads (the way that check was originally written)? I don't see how that'd be a race (it's read-read), but that _was_ what it did...

PR #123, opened 2020-04-09
Track more mut locations

jonhoo:


src/rt/atomic.rs
```diff
@@ -313,6 +313,9 @@ impl<T: Numeric> Atomic<T> {
                     state.stores[index].value = T::into_u64(self.0);
 
                     if !std::thread::panicking() {
+                        state
```
carllerche, line 316: Does this work? Afaik, it will only show the internal lib code as the location and will overwrite the "entry" point which is more useufl.

PR #127, opened 2020-04-11
Remove duplicated code on RwLock

pickfire:
@pop Thanks a lot for sending in the preview patch, I learned a lot there.


pop:
Thanks for fixing this. That duplicate code was really bothering me. It looks like a nice solution!

PR #138, opened 2020-05-01
Implement Default for Mutex, RwLock and Condvar

faern:
I noticed the `loom` versions of `Mutex` and `RwLock` does not implement `Default`. This PR adds that, so they become more similar.


faern:
I also added the `Default` impl for `Condvar` since that's another thing that was different from the standard library. Also rebased on latest master.

LucioFranco:
anything blocking this from being merged?

PR #140, opened 2020-05-01
Fix RwLock implementation

carllerche:
Fixes a RwLock bug where the reader and writer threads were not
correctly enabled / disabled in order to respect the RwLock guarantees.
This resulted in `write()` locks being granted while there are
outstanding readers.

cc @vakaras @pop

### DIFF
--- a/src/rt/rwlock.rs
+++ b/src/rt/rwlock.rs
@@ -54,7 +54,8 @@ impl RwLock {
     /// Acquire the read lock.
     /// Fail to acquire read lock if already *write* locked.
     pub(crate) fn acquire_read_lock(&self) {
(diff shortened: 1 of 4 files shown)


src/rt/rwlock.rs
```diff
@@ -111,25 +133,6 @@ impl RwLock {
         });
     }
 
-    fn lock_out_threads(&self, execution: &mut Execution, thread_id: thread::Id) {
```
carllerche, line 114: The fix required changing the fn details for each call... I opted to re-add some duplication instead of figuring out a way to stay DRY. I didn't think that adding another argument to the fn would be ideal.

PR #150, opened 2020-06-16
Add configuration option for max stack size

jamesbornholt:
I have some Loom tests that have started segfaulting because they're overflowing the stack created by generator (the tests are running real code, so I don't have a lot of control over how it uses the stack). generator has an option to pass in a stack size, but Loom doesn't currently expose it. This PR exposes that option in the Builder and the `LOOM_MAX_STACK_SIZE` environment variable, defaulting to 0x1000, which is generator's current default. One oddity of the API is that the stack size is specified in number of machine words rather than bytes, so 0x1000 actually specifies a 32KiB stack.


taiki-e:
Closing in favor of #311. Thanks anyway for the PR.

PR #151, opened 2020-06-20
Show features (futures) in docs

faern:
The fact that there is a `futures` feature flag and a `block_on` method was a bit hard to discover. At least for me. That lead me to create #148. Turns out it was not needed.

To make it more discoverable, this PR makes docs.rs build the documentation with all features active, so the extra functionality they enable are visible. It also uses the unstable `doc_cfg` attribute to make it more obvious in the docs that those items are feature gated.

This is taken more or less directly from how the tokio crate does the same thing.


hawkw:
It would be really nice to get this merged, but it looks like the PR is quite out of date. I'm going to update it myself.

PR #153, opened 2020-06-30
update scoped-tls to 1.0.0

Keruspe:

PR #155, opened 2020-07-06
Update `scoped-tls` to `1`

LucioFranco:


taiki-e:
seems dup of #153

LucioFranco:
It is!!! Thanks @taiki-e  :)

PR #161, opened 2020-07-26
Add support for lock_api

mystor:
Adds Loom mock implementations of the `lock_api::RawMutex` and `lock_api::RawRwLock` traits, as well as re-exports for the relevant concrete types, gated behind a `"lock_api"` feature.

The `parking_lot` and `lock_api` crates provide a slightly different interface than the standard library interface to `Mutex` and `RwLock`, which can make them difficult or impossible to test using `sync::Mutex` and `sync::RwLock` mocks.

### DIFF
--- a/src/lock_api.rs
+++ b/src/lock_api.rs
@@ -1,0 +1,114 @@
+//! Mock implementation of the `lock_api` and `parking_lot` crates.
+//!
+//! _These types are only available if loom is built with the `"lock_api"`
+//! feature_
+
+use crate::rt;
+use lock_api_ as lock_api;
+use once_cell::sync::OnceCell;
+
+/// Mock implementation of `lock_api::RawMutex`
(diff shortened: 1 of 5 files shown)


src/lock_api.rs
```diff
@@ -0,0 +1,123 @@
...
+    object: OnceCell<rt::Mutex>,
+}
+
+impl RawMutex {
+    // Unfortunately, we're required to have a `const INIT` in order to
+    // implement `lock_api::RawMutex`, so we need to lazily create the actual
+    // `rt::Mutex`.
+    fn object(&self) -> &rt::Mutex {
+        self.object.get_or_init(|| rt::Mutex::new(true))
+    }
```
mystor, line 22: I couldn't think of a cleaner way to satisfy this requirement, so I threw in a OnceCell.

PR #167, opened 2020-08-26
Adding `wait_while` to the loom Condvar. 

johnhurt:
This is just some logic combined with the existing wait, so I lifted the actual wait_while matches exactly what the std::rust implementation does

### DIFF
--- a/src/sync/condvar.rs
+++ b/src/sync/condvar.rs
@@ -38,6 +38,22 @@ impl Condvar {
         Ok(guard)
     }
 
+    /// Blocks the current thread until this condition variable receives a
+    /// notification and the given condition returns false.
+    pub fn wait_while<'a, T, F>(
+        &self,
+        mut guard: MutexGuard<'a, T>,
+        mut condition: F,
+    ) -> LockResult<MutexGuard<'a, T>>
+    where
+        F: FnMut(&mut T) -> bool,
+    {
+        while condition(&mut *guard) {
+            guard = self.wait(guard)?;
+        }
+        Ok(guard)
+    }
+
     /// Waits on this condition variable for a notification, timing out after a
     /// specified duration.
     pub fn wait_timeout<'a, T>(
--- a/tests/condvar.rs
+++ b/tests/condvar.rs
@@ -42,6 +42,22 @@ fn notify_all() {
     });
 }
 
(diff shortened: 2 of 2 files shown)


src/sync/condvar.rs
```diff
@@ -38,6 +38,22 @@ impl Condvar {
         Ok(guard)
     }
 
+    /// Blocks the current thread until this condition variable receives a
+    /// notification and the given condition returns false.
```
hawkw, line 42: Nit, take it or leave it: we _could_ just copy and paste the [entire doc comment from `std::sync::Condvar`](https://doc.rust-lang.org/stable/std/sync/struct.Condvar.html#method.wait_while) here, since the API is identical. But, it doesn't really matter (and it looks like other methods on our mock `Condvar` don't reproduce the entire `std` comment...

hawkw (approved):
looks good to me, thanks!

src/sync/condvar.rs
johnhurt: I thought about that, but yeah. It didn't seem like that would fit in with the existing comments that were there.

src/sync/condvar.rs
```diff
@@ -38,6 +38,22 @@ impl Condvar {
...
+    pub fn wait_while<'a, T, F>(
+        &self,
+        mut guard: MutexGuard<'a, T>,
+        mut condition: F,
+    ) -> LockResult<MutexGuard<'a, T>>
```
LucioFranco, line 47: I wonder if we could elide these?

LucioFranco (approved):

src/sync/condvar.rs
johnhurt: No, I tried. Eliding the lifetime completely gives a warning because it’s part of the MutexGuard struct. Trying to make it anonymous gives a complete error because it tries to use the same lifetime as &self. 

Also messing with signature seems like a bad idea since we’re trying match what’s in std.

emakryo:
Are there any blocking issues for this PR? I appreciate that we can use this API in loom. Thank you!

manforowicz:
I was going to submit a similar PR, but realized this one already exists. Any reason why it wasn't merged yet?

PR #172, opened 2020-09-11
Re-enable strong_count (possibly naively)

### LINKED ISSUE #171: Arc::strong_count is unimplemented
For some reason, `Arc::strong_count` is unimplemented with the following panic message given if it is called:

```
no tests checking this? DELETED!
```

It appears that this was [remarked on](https://github.com/tokio-rs/loom/pull/73#discussion_r331731979) in #73, but not addressed. Is there a reason for this function being missing?

Restioson:
This is a possibly naive, but if correct, would close #171.

### DIFF
--- a/src/sync/arc.rs
+++ b/src/sync/arc.rs
@@ -29,9 +29,8 @@ impl<T> Arc<T> {
     }
 
     /// Gets the number of strong (`Arc`) pointers to this value.
-    pub fn strong_count(_this: &Self) -> usize {
-        unimplemented!("no tests checking this? DELETED!")
-        // this.inner.ref_cnt.load(SeqCst)
+    pub fn strong_count(this: &Self) -> usize {
+        std::sync::Arc::strong_count(&this.inner)
     }
 
     /// Returns a mutable reference to the inner value, if there are


src/sync/arc.rs
```diff
@@ -29,9 +29,8 @@ impl<T> Arc<T> {
     }
 
     /// Gets the number of strong (`Arc`) pointers to this value.
-    pub fn strong_count(_this: &Self) -> usize {
-        unimplemented!("no tests checking this? DELETED!")
-        // this.inner.ref_cnt.load(SeqCst)
+    pub fn strong_count(this: &Self) -> usize {
+        std::sync::Arc::strong_count(&this.inner)
```
hawkw, line 33: I think this should return the value of the `Arc`'s modeled state's `ref_cnt` field: https://github.com/tokio-rs/loom/blob/9cd2cd21f015d2582308c705fe0e25c475e83cbb/src/rt/arc.rs#L14

And, I think this should synchronize the `Arc`'s state in the model with a `SeqCst` load, similarly to the other ref count ops: https://github.com/tokio-rs/loom/blob/73e6c14da41978e7a20e1993c2554a16a3dcb597/src/rt/arc.rs#L80-L81

That way, the returned value is actually part of the model, rather than bypassing it. In practice, I don't know if this is terribly important, but I think it's more correct.

hawkw (changes requested):
Thanks for working on this! IMO, it would be better to actually use the modeled state for the simulated `Arc`, rather than the ref count of the inner `std::sync::Arc`. 

@carllerche, what do you think?

Restioson:
Thanks for the feedback - I can look at this tomorrow again :)

src/sync/arc.rs
carllerche: In practice, this is going to be a bit more complicated.

`Arc` implements its own "dependent access" tracking in order to be more efficient about reducing the state space. You will see in `rt/arc.rs` there is `enum Action` and this matters for the implementations of `last_dependent_access` and `set_last_access`. To implement `strong_count` correctly, we would need to update the type to manage the dependent access correctly.

The way to think about it is that two concurrent increments are not dependent, i.e. the order in which they execute does not impact the behavior of the program, so loom doesn't bother permuting it.

I believe a `strong_count` call is dependent with both incs and decs, but not other calls to `strong_count()`.

Restioson: Okay, thank you for the information :) I do not really know how loom works internally (I just began using it) so I will try find my way around it based on what's been said here.

Restioson:
I've made an attempt to implement the suggestions given above, but I'm not quite sure how to test that this works.

mccolljr:
Don't mean to hijack, but are there plans to push this forward? Anything I can help with here? Would love to make use of this in a test I'm writing

Restioson:
> Don't mean to hijack, but are there plans to push this forward? Anything I can help with here? Would love to make use of this in a test I'm writing

From my side, I am waiting for someone to review it and make sure that it makes sense, as I'm unsure exactly of how loom works internally.

hawkw (approved):
The new implementation (as of 170d271c18e8f070ff1677626efcfbf727955c26) looks correct to me, thanks!

It would be nice to get an approval from @carllerche as well.

hawkw:
Regarding:

> I've made an attempt to implement the suggestions given above, but I'm not quite sure how to test that this works.

I think it would, at least, be nice to have a test where we have a `loom::sync::Arc` and wrap it in some type that implements identical reference counting with an `AtomicUsize` --- so that when this type is cloned, both the `Arc`'s reference count and the reference count in the `AtomicUsize` are incremented, and when the type is dropped, the `Arc`'s ref count and the `AtomicUsize's ref count is decremented. Then, we can assert that all threads observe equivalent values for the `Arc` and the `AtomicUsize`. 

Note that I think we would actually want to use `std::sync::AtomicUsize`, _not_ the `loom` `AtomicUsize`. This is so that the increments and decrements to the `AtomicUsize` are _not_ treated as separate ops in the simulation. If we use a `loom` `AtomicUsize` and a `loom` `Arc` in the same test, then loom could (quite correctly) note that it's possible for a thread to (say) increment the `Arc``'s ref count and then get preempted immediately before it increments the `AtomicUsize` reference count, resulting in a discrepancy between the two. In a test for `strong_count`, though, we would want to keep the two in sync at all times, with the `AtomicUsize` (outside of loom) representing the "actual" expected reference count.

Restioson:
Doesn't this make it go into instant overtime by changing the amount of time to 0?

Restioson:
Hi, thanks for the feedback! I can take a look at this over the weekend

hawkw:
> Hi, thanks for the feedback! I can take a look at this over the weekend

Hi @Restioson, have you had a chance  to look at testing this? If not...we can always merge this PR as-is, and add tests later.

hawkw (approved):
I think it's probably fine to go ahead and merge this one without waiting to add tests --- @carllerche, any last concerns before I merge this?

hawkw (changes requested):
oh, this no longer compiles with recent changes to `master`, i'm going to go ahead and fix it up

hawkw (approved):
Okay, now that this builds successfully against the latest `master`, I _think_ this looks correct. @carllerche, do you want to give it one last look?

PR #189, opened 2020-12-03
Prepare 0.4 and add I variant atomic ints

LucioFranco:


carllerche (approved):
Probably want to remove `publish = false`

PR #195, opened 2021-03-13
Fix compare_and_swap deprecation warning

thomcc:
Trivial patch, but I have two other patches incoming, and it didn't really make sense to do it in either of them.


jonhoo:
Oops, I fixed this in https://github.com/tokio-rs/loom/pull/201 and missed this! In any case, this is now fixed :)

PR #196, opened 2021-03-13
Add a loom::cell::Cell, which provides a basic wrapper of the loom UnsafeCell

thomcc:
It isn't too rare to use `Cell<T>` instead of `UnsafeCell<T>` if `T` is `Copy`. It also is way more compatible with the stdlib API. Wasn't sure what to do for tests here.


thomcc:
Stable CI failure is a warning that's fixed by #195

Darksonn (approved):

PR #197, opened 2021-03-13
Add a `loom::hint` module containing mocked versions of `spin_loop` and `unreachable_unchecked`

thomcc:
`std::sync::atomic::spin_loop_hint` is deprecated, so loom should provide the non-deprecated replacement. It just calls the deprecated version though.

I provided `unreachable_unchecked` too, even though it's not *exactly* `loom`'s wheelhouse, but it could help catch some UB, and lets users potentially reduce the amount of `cfg` special-casing they need.

### DIFF
--- a/src/hint.rs
+++ b/src/hint.rs
@@ -1,0 +1,15 @@
+//! Mocked versions of [`std::hint`] functions.
+
+/// Signals the processor that it is entering a busy-wait spin-loop.
+pub fn spin_loop() {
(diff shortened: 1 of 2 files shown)


thomcc:
Stable CI failure is a warning that's fixed by #195.

carllerche (approved):

carllerche:
Thanks :+1:

PR #199, opened 2021-04-01
Revamp the documentation

jonhoo:


carllerche (approved):

src/lib.rs
```diff
@@ -74,49 +90,242 @@
...
 //!
 //! # Writing tests
 //!
-//! Test cases using loom must be fully determinstic. All sources of
-//! non-determism must be via loom types. This allows loom to validate the test
-//! case and control the scheduling.
+//! Test cases using loom must be fully determinstic. All sources of non-determism must be via loom
+//! types so that loom can expose different possible values on each execution of the test closure.
+//! Other sources of non-determinism like random number generation or system calls cannot be
+//! modeled by loom.
```
carllerche, line 103: Might be worth calling out that one can mock out things like random number generation or syscalls when running a loom test. The Tokio runtime does this to use loom.

src/lib.rs
```diff
@@ -74,49 +90,242 @@
...
-//! assumes that, at some point, another thread will make enough progress for
-//! the lock to become available.
+//! ## Yielding
 //!
-//! This presents a challenge for loom as the scheduler is not fair. In such
-//! cases, loops must include calls to `yield_now`. This tells loom that another
-//! thread needs to be scheduled in order for the current one to make progress.
+//! Some concurrent algorithms assume a fair scheduler. For example, a spin lock assumes that, at
+//! some point, another thread will make enough progress for the lock to become available. This
+//! presents a challenge for loom as the scheduler is not fair. In such cases, loops must include
```
carllerche, line 171: You could expand that, technically, the spinning thread doesn't ever have to yield back to the scheduler so loom will explore that option and blow the stack. This is also a good reason to not use spin locks :)

jonhoo: And also to link to matklad's excellent https://matklad.github.io/2020/01/02/spinlocks-considered-harmful.html

src/lib.rs
```diff
@@ -74,49 +90,242 @@
...
+//! ```
+//!
+//! Unfortunately, it is not possible for loom to completely model all the interleavings that
+//! relaxed memory ordering allows. This is because the relaxed memory ordering allows memory
+//! operations to be re-ordered within a single thread -- B can run *before* A -- which loom cannot
+//! emulate. The same restriction applies to certain reorderings that are possible across different
+//! atomic variables with other memory orderings.
+//!
+//! Ultimately, this restriction means that loom is not a _completely_ exhaustive checker, and may
+//! miss bugs that could occur under particularly weak memory models.
```
carllerche, line 302: > miss bugs that could occur under particularly weak memory models.

This is not really clear. Loom checks the C++ memory model, which holds across all CPUs (that implement it). I'm *guessing* you are trying to say that loom may not catch bugs that can happen on ARM (CPU memory model), but this specific bug is probably an artifact of a compiler optimization and not the CPU's memory model.

Also, if one does not use `Relaxed`, then loom should be complete (modulo bugs, of which we know of one w/ SeqCst).

src/lib.rs
```diff
@@ -74,49 +90,242 @@
...
+//!     LOOM_CHECKPOINT_FILE=my_test.json \
+//!     RUSTFLAGS="--cfg loom" \
+//!     [other env vars] \
+//!     cargo test --test loom_my_struct --release [failing test]
+//! ```
+//!
+//! This should provide you with a trace of all the concurrency events leading up to the failure,
+//! which should allow you to identify how the bug is triggered.
+//!
+//! # Limitations and Caveats
```
carllerche, line 249: It's also probably worth adding a section on loom being an "intrusive" tool requiring the user to use loom's APIs. The implication being that loom cannot catch bugs that are from code *not* instrumented. `tsan` does not have this problem as it is not intrusive.

src/lib.rs
jonhoo: The memory order affects both what the CPU is allowed to do _and_ what the compiler is allowed to do, so in this case I'm trying to get at the fact that the memory model allows a relaxed operations to appear to happen in different orders on different threads. How exactly that happens isn't all that important.

I think the same fundamental limitation comes up with `Acquire`+`Release` too in certain cases. Specifically, if a thread does a `store(Release)` on one variable followed by a `load(Acquire)` on another variable, I _think_ it is legal for the compiler or CPU to expose them in the opposite order to other threads, which I don't think loom can model the effects of?

jonhoo: I ended up removing that last short paragraph. I think the paragraph above is sufficient.

PR #200, opened 2021-04-01
Add C++ memory ordering spec tests

jonhoo:
https://en.cppreference.com/w/cpp/atomic/memory_order


jonhoo:
The `Relaxed` test fails due to known limitations, but I figured it should be included for completeness.

The `SeqCst` test fails due to a bug that should be fixed, but I haven't had a chance to investigate yet. It's probably the same as https://github.com/tokio-rs/loom/issues/180.

carllerche (approved):

PR #203, opened 2021-04-02
Update generator to 0.7

taiki-e:
Fixes #198

PR #205, opened 2021-04-02
Bump MSRV to 1.51

taiki-e:
generator 0.7 requires Rust 1.51: https://github.com/tokio-rs/loom/pull/203#issuecomment-812500571

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,7 +39,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Install Rust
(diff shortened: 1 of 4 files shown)


carllerche:
I'm fine w/ this, but maybe we can also bump loom to v0.5 as part of it? cc/ @jonhoo 

If yes, could you change the version in Cargo.toml and set publish = false?

jonhoo:
Yeah, given that we're still <1.0 I think bumping the minor makes sense here for the MSRV bump.

taiki-e:
Bumped loom to 0.5.0 and marked as 'publish = false'.

carllerche (approved):

PR #207, opened 2021-04-02
Fix link to 'C11 Memory Model'

jonhoo:


taiki-e (approved):

jonhoo:
Actually, is there a better link for this than to cppreference? Is there a reason we specifically call out C11 as opposed to other versions or the C++ version? @carllerche

carllerche:
Not sure. What’s the difference?

jonhoo:
I don't actually have the diffs, but from cppreference.com it's clear that some of the specification has changed across the different versions of the C++ specification. For example, the formal specification for sequential consistency changed in C++20. I'm also guessing that the C and C++ versions are _slightly_ different, even if the underlying memory model is probably the same. Are we _specifically_ matching C11 (or C++11?), or is the expectation that loom matches "the current C/C++ specification, whatever it is"? The page we currently link to implies "the current C++ spec", even though the _text_ says "C11".

As for other things we could link to, we could include references to the relevant sections on memory ordering from latest C and C++ specs (which aren't available for free, although the drafts which are basically the same are). I think the cppreference link is probably a good one, since it mostly just extracts what's in the spec, but it's a sort of "informal" reference (and again, the C11 label isn't _really_ accurate).

PR #210, opened 2021-04-03
Add From implementation to atomic types

### LINKED ISSUE #194: Cannot convert numbers `Into` atomic integers
I have code which tries to convert a numeric literal into a Loom 0.4.0 atomic integer (`let foo: AtomicU8 = 1.into()`). This works with std, but with Loom types, I get the following error:

> the trait bound `loom::sync::atomic::AtomicU8: From<{integer}>` is not satisfied
> required because of the requirements on the impl of `Into<loom::sync::atomic::AtomicU8>` for `{integer}`

I think there's a missing trait conversion (either From or Into, not sure which).

Loom may also be missing the ability to convert atomics into bare integers (didn't test if it works on std).

taiki-e:
std atomic types provide them: https://doc.rust-lang.org/nightly/core/sync/atomic/struct.AtomicU8.html#impl-From%3Cu8%3E

Closes #194

### DIFF
--- a/src/sync/atomic/bool.rs
+++ b/src/sync/atomic/bool.rs
@@ -97,3 +97,9 @@ impl Default for AtomicBool {
         AtomicBool::new(Default::default())
     }
 }
+
+impl From<bool> for AtomicBool {
+    fn from(b: bool) -> Self {
+        Self::new(b)
+    }
+}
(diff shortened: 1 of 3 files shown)


carllerche (approved):
Nice!

PR #221, opened 2021-08-31
Implement try_recv function for mpsc Receiver.

gz:
I'm not familiar enough with loom to judge if this is the right way to do it, but at least it seems to be working in my tests.


hawkw (approved):
I think this seems correct to me.

PR #222, opened 2021-09-07
Remove implicit `T: Sized` requirement from `UnsafeCell`

WaffleLapkin:
`std`'s `UnsafeCell`﻿ doesn't have `T: Sized` requirement, so I don't see why the `loom` version should be different.


hawkw (approved):
looks good to me!

i think the missing `?Sized` bound may have been an artifact of a previous implementation where the data was not stored in a `std` `UnsafeCell`, but I may be wrong about that. it could have just been an oversight...

PR #226, opened 2021-09-14
Remove (implicit) `T: Sized` bound from `Arc`

### LINKED ISSUE #85: Arc requires T: Sized
`std::sync::Arc<T>` does not require `T: Sized`, where as loom's version does. This means that code that uses, say, `Arc<dyn Fn>` will not compile with loom, which is unfortunate. I _believe_ we would need to move the `T` to the end of `loom::Arc` for this to work, though that also breaks the layout needed according to `loom::Arc`'s comment:
https://github.com/carllerche/loom/blob/339dd48f57d3b9b786e44db4914d08882d7659bf/src/sync/arc.rs#L11-L17

WaffleLapkin:
This is an attempt to resolve #85

### DIFF
--- a/src/sync/arc.rs
+++ b/src/sync/arc.rs
@@ -31,9 +31,9 @@ impl<T> Arc<T> {
 }
 
 impl<T: ?Sized> Arc<T> {
-    /// Converts `std::sync::Arc` to `loom::sync::Arc`
+    /// Converts `std::sync::Arc` to `loom::sync::Arc`.
     ///
-    /// This is needed to create a `loom::sync::Arc<T>` where `T: !Sized`
+    /// This is needed to create a `loom::sync::Arc<T>` where `T: !Sized`.
     ///
     /// ## Panics
     ///
@@ -41,16 +41,36 @@ impl<T: ?Sized> Arc<T> {
     ///
     /// ## Examples
     ///
+    /// While `std::sync::Arc` with `T: !Sized` can be created by coercing an
+    /// `std::sync::Arc` with a sized value:
+    ///
+    /// ```rust
+    /// let sized: std::sync::Arc<[u8; 3]> = std::sync::Arc::new([1, 2, 3]);
+    /// let _unsized: std::sync::Arc<[u8]> = sized; // coercion
+    /// ```
+    ///
+    /// `loom::sync::Arc` can't be created in the same way:
+    ///
+    /// ```compile_fail,E0308
+    /// use loom::sync::Arc;
+    ///
+    /// let sized: Arc<[u8; 3]> = Arc::new([1, 2, 3]);
+    /// let _unsized: Arc<[u8]> = sized; // error: mismatched types
+    /// ```
+    ///
+    /// This is because `std::sync::Arc` uses an unstable trait called `CoerceUnsized`
+    /// that loom can't use. To create `loom::sync::Arc` with an unsized inner value
+    /// first create a `std::sync::Arc` of an appropriate type and then use this method:
+    ///
     /// ```rust
     /// use loom::sync::Arc;
     ///
     /// # loom::model::model(|| {
-    /// // std's arc can be automatically coerced
-    /// let std = std::sync::Arc::new([1, 2, 3]);
+    /// let std: std::sync::Arc<[u8]> = std::sync::Arc::new([1, 2, 3]);
     /// let loom: Arc<[u8]> = Arc::from_std(std);
     ///
-    /// let std = std::sync::Arc::new([1, 2, 3]);
-    /// let loom: Arc<dyn Sync + Send> = Arc::from_std(std);
+    /// let std: std::sync::Arc<dyn Send + Sync> = std::sync::Arc::new([1, 2, 3]);
+    /// let loom: Arc<dyn Send + Sync> = Arc::from_std(std);
     /// # });
     /// ```
     #[track_caller]
--- a/src/sync/arc.rs
+++ b/src/sync/arc.rs
@@ -33,7 +33,7 @@ impl<T> Arc<T> {
 impl<T: ?Sized> Arc<T> {
     /// Converts `std::sync::Arc` to `loom::sync::Arc`
     ///
-    /// This is needed to create `Arc<T>` where `T: !Sized`
+    /// This is needed to create a `loom::sync::Arc<T>` where `T: !Sized`
     ///
     /// ## Panics
     ///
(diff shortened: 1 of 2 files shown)


hawkw:
Sorry this took so long to review --- this seems like a reasonable approach! @WaffleLapkin are you interested in continuing to work on this?

WaffleLapkin:
Yes, I'm still interested!

WaffleLapkin:
@hawkw I believe this is ready for review

WaffleLapkin:
(@hawkw friendly ping)

src/sync/arc.rs
```diff
@@ -5,36 +5,72 @@ use std::{mem, ops};
...
+}
+
+impl<T: ?Sized> Arc<T> {
+    /// Converts `std::sync::Arc` to `loom::sync::Arc`
+    ///
+    /// This is needed to create `Arc<T>` where `T: !Sized`
+    ///
+    /// ## Panics
+    ///
+    /// If the provided arc has copies (ie if it is not unique).
```
hawkw, line 40: couple nits:
```suggestion
    /// If the provided `Arc` has copies (i.e., if it is not unique).
```

src/sync/arc.rs
```diff
@@ -5,36 +5,72 @@ use std::{mem, ops};
...
 
     /// Constructs a new `Pin<Arc<T>>`.
     pub fn pin(data: T) -> Pin<Arc<T>> {
         unsafe { Pin::new_unchecked(Arc::new(data)) }
     }
 
+    /// Returns the inner value, if the `Arc` has exactly one strong reference.
+    pub fn try_unwrap(_this: Arc<T>) -> Result<T, Arc<T>> {
+        unimplemented!();
+    }
```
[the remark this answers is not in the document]
WaffleLapkin: I agree. This PR just moves this function ([from here](https://github.com/tokio-rs/loom/pull/226/files#diff-f76b60530ef5a0519d08adea1a00553e7e8daa7614dde8438c7ed39e3b7ceae4L128)). I think we can remove/properly implement it in a separate PR.

src/sync/arc.rs
```diff
@@ -5,36 +5,72 @@ use std::{mem, ops};
...
+}
+
+impl<T: ?Sized> Arc<T> {
+    /// Converts `std::sync::Arc` to `loom::sync::Arc`
+    ///
+    /// This is needed to create `Arc<T>` where `T: !Sized`
```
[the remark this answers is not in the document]
WaffleLapkin: `loom::sync::Arc::new` accepts the argument by value (just as `std::sync::Arc::new`) and so it can't accept `?Sized` types.

The same can be said about `std::sync::Arc::new`, it too can only be called with `T: Sized`. The magic, that allows creating `std::sync::Arc` with `!Sized` types is "unsize-coercion" via [`CoerceUnsized`](https://doc.rust-lang.org/std/sync/struct.Arc.html#impl-CoerceUnsized%3CArc%3CU%3E%3E).

carllerche:
@hawkw what is left for this PR?

src/sync/arc.rs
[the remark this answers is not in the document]
hawkw: ah, yeah, that makes sense. thanks!

src/sync/arc.rs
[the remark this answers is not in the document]
hawkw: ah, i missed that this was already there, my bad. :+1:

src/sync/arc.rs
hawkw, line 36: minor docs nit: maybe
```suggestion
    /// This is needed to create a `loom::sync::Arc<T>` where `T: !Sized`
```

src/sync/arc.rs
```diff
@@ -5,36 +5,72 @@ use std::{mem, ops};
...
-#[derive(Debug)]
-#[repr(C)]
-struct Inner<T> {
-    // This must be the first field to make into_raw / from_raw work
-    value: T,
-
-    obj: rt::Arc,
+pub struct Arc<T: ?Sized> {
+    obj: std::sync::Arc<rt::Arc>,
+    inner: std::sync::Arc<T>,
```
hawkw, line 10: nit, take it or leave it: should we maybe consider renaming this field from `inner` to `value` or something? since it now stores only the actual `Arc`ed value, it might be worth making that clearer in the `Debug` output? I note that the field holding the Arc's value used to be `inner.value`...

src/sync/arc.rs
```diff
@@ -5,36 +5,72 @@ use std::{mem, ops};
...
+    ///
+    /// If the provided `Arc` has copies (i.e., if it is not unique).
+    ///
+    /// ## Examples
+    ///
+    /// ```rust
+    /// use loom::sync::Arc;
+    ///
+    /// # loom::model::model(|| {
+    /// // std's arc can be automatically coerced
```
hawkw, line 48: nit: maybe
```suggestion
    /// // std's arc can be automatically coerced from a `!Sized` value
```

src/sync/arc.rs
```diff
@@ -5,36 +5,72 @@ use std::{mem, ops};
...
+    /// let std = std::sync::Arc::new([1, 2, 3]);
+    /// let loom: Arc<[u8]> = Arc::from_std(std);
+    ///
+    /// let std = std::sync::Arc::new([1, 2, 3]);
+    /// let loom: Arc<dyn Sync + Send> = Arc::from_std(std);
+    /// # });
```
hawkw, line 54: a very minor docs suggestion, take it or leave it: it could also be nice to have a `compile_fail` example showing how `loom::sync::Arc::new` _cannot_ be used with `!Sized` values directly, so that users can see why this function is necessary? not a blocker.

hawkw (approved):
overall, this seems great to me --- i left some very minor nitpicky comments, but I think this looks good!

src/sync/arc.rs
WaffleLapkin: That's not really correct though, the coercion goes `Arc<[u8; 3]>` -> `Arc<[u8]>`, so it is not "from a `!Sized` value" but more like "to a `!Sized` value"?

WaffleLapkin:
@hawkw I believe I've addressed your nitpicks :)

src/sync/arc.rs
hawkw: ah, okay, "to" would also be fine, i just thought it was nice if the comment explicitly stated the coercion that was happening. not a big deal!

hawkw (approved):
docs changes look great, thanks!

WaffleLapkin:
Yay! 🎉

PR #238, opened 2021-11-13
Update tracing-subscriber to 0.3

### LINKED ISSUE #237: Remove abandoned chrono library from dependencies
[chrono](https://crates.io/crates/chrono) is a library that is abandoned and has [unresolved](https://github.com/chronotope/chrono/issues/499) security issues. As such, it is undesirable to have in the dependency closure.

Loom pulls in chrono by depending on `tracing-subscriber = "0.2.22"`, which includes chrono. Newer versions of `tracing-subscriber` (0.3.x) do not include this dependency. As such, I would propose to bump this dependency to remove the zombie chrono library from the dependencies. :)

taiki-e:
I guess this unblocks https://github.com/tokio-rs/tokio/pull/4227

Closes #237

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,7 @@ serde_json = { version = "1.0.33", optional = true }
 pin-utils = { version = "0.1.0", optional = true }
 
 tracing = "0.1.27"
-tracing-subscriber = "0.2.22"
+tracing-subscriber = "0.3"
 
 [dev-dependencies]
 futures-util = "0.3.0"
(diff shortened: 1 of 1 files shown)


hawkw (approved):
Oh, _that_'s why 0.2 was still in `tokio`'s deps, but I wasn't seeing it in `cargo tree` because of the `cfg` attribute... in hindsight, that seems pretty obvious. Nice catch!

I believe this will also fix #237

PR #239, opened 2021-11-13
fix a bunch of annoying clippy lints

hawkw:
This branch fixes a big pile of clippy lints. These are mostly not
particularly important, but they were kinda bugging me, so I thought it
was nice to fix them.

I've tried to separate out each individual lint category into separate
commits, with notes in the commit message as necessary. Most of these
changes are completely trivial, but there are a few that might be worth
actually reviewing in detail. Therefore, each commit can be considered
individually, and if there are any cases where we don't like Clippy's
suggestions, I can back out those commits and add `#[allow(...)]`
attributes instead.

In particular, `clippy` suggested adding
`Default` implementations for a few types. This seems pretty harmless,
but (unlike the rest of this PR) it *is* an actual API change, so if
that's unwanted, I can back out that commit and add `allow` attributes
for the `new_without_default` lint, instead.

I've also added some new documentation to resolve the
`clippy::missing_safety_doc` lint. In most cases, this was on functions
that mimic `std` APIs that are unsafe, so I've copied the `# Safety`
sections from the `std` docs for those functions. In some cases I've
added additional text to the documentation describing differences
between `loom`'s version and `std`'s versions of these functions.

### DIFF
--- a/src/alloc.rs
+++ b/src/alloc.rs
@@ -5,6 +5,20 @@ use crate::rt;
 pub use std::alloc::Layout;
 
 /// Allocate memory with the global allocator.
+///
+/// This is equivalent to the standard library's [`std::alloc::alloc`], but with
+/// the addition of leak tracking for allocated objects. Loom's leak tracking
+/// will not function for allocations not performed via this method.
+///
+/// This function forwards calls to the [`GlobalAlloc::alloc`] method
+/// of the allocator registered with the `#[global_allocator]` attribute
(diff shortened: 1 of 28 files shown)


Noah-Kennedy (approved):
This looks good.

hawkw:
cc @carllerche any concerns about the API additions, or shall i just go ahead and merge this?

Darksonn (approved):

PR #250, opened 2022-01-06
thread: fix semantics of `thread::park` after `Thread::unpark`

### LINKED ISSUE #246: incorrect semantics for `Thread::unpark` followed by `thread::park`
If `std::thread::Thread::unpark` is called _prior_ to a call to `std::thread::park`, the `park` call will return immediately, "consuming" the previous unpark. Per [the documentation](https://doc.rust-lang.org/std/thread/fn.park.html#park-and-unpark):

> The `unpark` method on a `Thread` atomically makes the token available if it wasn’t already. Because the token is initially absent, `unpark` followed by `park` will result in the second call returning immediately.

Loom's version of this API does not have this behavior. It appears that `unpark` only notifies a thread if it is currently `park`ed, and does not influence the next call to `park`.

The following test demonstrates this:
```rust
#[test]
fn park_unpark_loom() {
    loom::model(|| {
        loom::thread::current().unpark();
        loom::thread::park();
        println!("it did not deadlock");
    });
}

#[test]
fn park_unpark_std() {
    std::thread::current().unpark();
    std::thread::park();
    println!("it did not deadlock");
}
```

The `park_unpark_std` test will run to completion and print "it did not deadlock", while the `park_unpark_loom` test panics with a deadlock detection error:

```
running 2 tests
test park_unpark_std ... ok
test park_unpark_loom ... FAILED

failures:

---- park_unpark_loom stdout ----
unpark
park
thread 'park_unpark_loom' panicked at 'deadlock; threads = [(Id(0), Blocked)]', src/rt/execution.rs:209:13
stack backtrace:
   ...
```

hawkw:
## Motivation

If `std::thread::Thread::unpark` is called _prior_ to a call to
`std::thread::park`, the `park` call will return immediately,
"consuming" the previous unpark. Per [the documentation][1]:

> The `unpark` method on a `Thread` atomically makes the token available
> if it wasn’t already. Because the token is initially absent, `unpark`
> followed by `park` will result in the second call returning
> immediately.

Loom's version of this API does not have this behavior. It appears that
`unpark` only notifies a thread if it is currently `park`ed, and does
not influence the next call to `park`.

The following test demonstrates this:

```rust
#[test]
fn park_unpark_loom() {
    loom::model(|| {
        loom::thread::current().unpark();
        loom::thread::park();
        println!("it did not deadlock");
    });
}

#[test]
fn park_unpark_std() {
    std::thread::current().unpark();
    std::thread::park();
    println!("it did not deadlock");
}
```

The `park_unpark_std` test will run to completion and print "it did not
deadlock", while the `park_unpark_loom` test panics with a deadlock
detection error:

```
running 2 tests
test park_unpark_std ... ok
test park_unpark_loom ... FAILED

failures:

---- park_unpark_loom stdout ----
unpark
park
thread 'park_unpark_loom' panicked at 'deadlock; threads = [(Id(0), Blocked)]', src/rt/execution.rs:209:13
stack backtrace:
   ...
```

See issue #246.

## Solution

This branch fixes the semantics of `loom::thread::Thread::unpark` to
match the behavior of the standard library. This is done by adding a
boolean to `rt::thread::State::Runnable` to track whether a thread was
unparked while it was runnable. When unparking a runnable thread, the
boolean is set; when parking, we check if the boolean is set, and unset
it, rather than actually setting the state to blocked.

This means that `unpark` followed by `park` should now behave the same
as `std::thread`.

Additionally, it was necessary to change `rt::thread::Set::unpark` a
little bit to handle threads unparking themselves. In the current
implementation, if a thread calls `unpark` on itself, nothing happens.
Now, though, we still have to update the thread's state to store the
unparking.

Fixes #246

[1]: https://doc.rust-lang.org/std/thread/fn.park.html#park-and-unpark

### DIFF
--- a/src/rt/mod.rs
+++ b/src/rt/mod.rs
@@ -82,17 +82,32 @@ where
 
 /// Marks the current thread as blocked
 pub fn park() {
-    execution(|execution| {
+    let switch = execution(|execution| {
+        use thread::State;
         let thread = execution.threads.active_id();
+        let active = execution.threads.active_mut();
 
-        trace!(?thread, "park");
+        trace!(?thread, ?active.state, "park");
+
+        match active.state {
+            // The thread was previously unparked while it was active. Instead
+            // of parking, consume the unpark.
+            State::Runnable { unparked: true } => {
+                active.set_runnable();
+                return false;
+            }
+            // The thread doesn't have a saved unpark; set its state to blocked.
+            _ => active.set_blocked(),
+        };
 
         execution.threads.active_mut().set_blocked();
         execution.threads.active_mut().operation = None;
         execution.schedule()
     });
 
-    Scheduler::switch();
+    if switch {
+        Scheduler::switch();
+    }
 }
 
 /// Add an execution branch point.
--- a/src/rt/thread.rs
+++ b/src/rt/thread.rs
@@ -67,7 +67,7 @@ impl Id {
 
 #[derive(Debug, Clone, Copy)]
 pub(crate) enum State {
-    Runnable,
+    Runnable { unparked: bool },
     Blocked,
     Yield,
     Terminated,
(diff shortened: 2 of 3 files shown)


carllerche (approved):

carllerche:
You want to do a release?

PR #255, opened 2022-03-30
Change derive(Debug) AtomicPtr to an explicit impl for all T, not just T: Derive.

zjijz:
I think the root issue here is one of that tasks related to [rust#26925](https://github.com/rust-lang/rust/issues/26925). `AtomicPtr<T>` can still have a debug implementation for when `T` is not itself `Debug`.

This was brought up from doing work on [haphazard#32](https://github.com/jonhoo/haphazard/pull/32).


Darksonn (approved):
Thanks.

PR #258, opened 2022-04-20
track locations

carllerche:
This PR is mostly just mechanical. It adds `#[track_caller]` and gets the location in the public API and plumbs it through. This helps debug deadlocks:

```
thread 'runtime::tests::loom_pool::group_a::pool_multi_spawn' panicked at 'deadlock; threads = [(Id(0), Blocked(Location(Some(Location { file: "tokio/src/runtime/tests/loom_oneshot.rs", line: 45, col: 31 })))), (Id(1), Blocked(Location(Some(Location { file: "tokio/src/park/thread.rs", line: 112, col: 30 })))), (Id(2), Blocked(Location(Some(Location { file: "tokio/src/runtime/thread_pool/park.rs", line: 206, col: 30 }))))]'
```

The output is pretty bad, but it is definitely helpful :)

### DIFF
--- a/src/sync/arc.rs
+++ b/src/sync/arc.rs
@@ -72,8 +72,9 @@ impl<T> Arc<T> {
 
     /// Returns a mutable reference to the inner value, if there are
     /// no other `Arc` pointers to the same value.
+    #[track_caller]
     pub fn get_mut(this: &mut Self) -> Option<&mut T> {
-        if this.inner.obj.get_mut() {
+        if this.inner.obj.get_mut(location!()) {
             assert_eq!(1, std::sync::Arc::strong_count(&this.inner));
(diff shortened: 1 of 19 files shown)


Noah-Kennedy (approved):

src/sync/arc.rs
```diff
@@ -140,7 +141,7 @@ impl<T> ops::Deref for Arc<T> {
 
 impl<T> Clone for Arc<T> {
     fn clone(&self) -> Arc<T> {
-        self.inner.obj.ref_inc();
+        self.inner.obj.ref_inc(location!());
```
hawkw, line 144: do you think we should add `#[track_caller]` attributes to the `Clone` and `Drop` impls, too, so that we record the location in the _user_ code where the `Arc` was cloned or dropped from?

PR #262, opened 2022-04-26
Implement Arc::try_unwrap

### LINKED ISSUE #261: Implement Arc::try_unwrap()
Currently a blocker for me since it's missing.

kvark:
Fixes #261 

I'm not sure if the unsafe destruction path is the best here.

### DIFF
--- a/src/sync/arc.rs
+++ b/src/sync/arc.rs
@@ -1,7 +1,7 @@
 use crate::rt;
 
 use std::pin::Pin;
-use std::{mem, ops};
+use std::{mem, ops, ptr};
 
 /// Mock implementation of `std::sync::Arc`.
 #[derive(Debug)]
@@ -25,8 +25,30 @@ impl<T> Arc<T> {
     }
 
     /// Returns the inner value, if the `Arc` has exactly one strong reference.
-    pub fn try_unwrap(_this: Arc<T>) -> Result<T, Arc<T>> {
-        unimplemented!();
+    pub fn try_unwrap(this: Arc<T>) -> Result<T, Arc<T>> {
+        if this.obj.get_mut(location!()) {
+            assert_eq!(1, std::sync::Arc::strong_count(&this.value));
+
+            // work around our inability to destruct the object normally,
+            // because of the `Drop` presense.
+            this.obj.ref_dec(location!());
+            this.unregister();
+
+            let arc_value = unsafe {
+                let _arc_obj = ptr::read(&this.obj);
+                let arc_value = ptr::read(&this.value);
+
+                mem::forget(this);
+
+                arc_value
+            };
+            match std::sync::Arc::try_unwrap(arc_value) {
+                Ok(value) => Ok(value),
+                Err(_) => unreachable!(),
+            }
+        } else {
+            Err(this)
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/sync/arc.rs
```diff
@@ -25,8 +25,30 @@ impl<T> Arc<T> {
     }
 
     /// Returns the inner value, if the `Arc` has exactly one strong reference.
-    pub fn try_unwrap(_this: Arc<T>) -> Result<T, Arc<T>> {
-        unimplemented!();
+    pub fn try_unwrap(this: Arc<T>) -> Result<T, Arc<T>> {
```
hawkw, line 28: this probably needs a
```suggestion
    #[track_caller]
    pub fn try_unwrap(this: Arc<T>) -> Result<T, Arc<T>> {
```
on it so that the `location!()` passed to `get_mut` has the callsite in the user code where `try_unwrap` was called.

src/sync/arc.rs
```diff
@@ -25,8 +25,30 @@ impl<T> Arc<T> {
...
+            this.unregister();
+
+            let arc_value = unsafe {
+                let _arc_obj = ptr::read(&this.obj);
+                let arc_value = ptr::read(&this.value);
+
+                mem::forget(this);
+
+                arc_value
+            };
```
hawkw, line 44: a potential safe alternative to this is to wrap the loom `Arc`'s value field in an `Option` so that we can `take` it out without dropping the struct...but, this would mean every other time the `Arc`'s value is accessed would have to `unwrap()` it, which might have performance implications and make tests slower...this might be worth the unsafe code to avoid that? 

it would be nice to add a comment explaining why this is safe, though.

src/sync/arc.rs
```diff
@@ -25,8 +25,30 @@ impl<T> Arc<T> {
     }
 
     /// Returns the inner value, if the `Arc` has exactly one strong reference.
-    pub fn try_unwrap(_this: Arc<T>) -> Result<T, Arc<T>> {
-        unimplemented!();
+    pub fn try_unwrap(this: Arc<T>) -> Result<T, Arc<T>> {
+        if this.obj.get_mut(location!()) {
```
hawkw, line 29: nit, take it or leave it: i have a personal preference for changing this from an `if/`else` to an `if` with an early return:
```suggestion
        if !this.obj.get_mut(location!()) {
             return Err(this);
        }
```
so that the rest of the code isn't indented inside of the `if` block...but, this is just a personal style preference.

src/sync/arc.rs
kvark: I don't have a strong preference. Current approach is borrowed from libstd. Added a comment now.

kvark:
please take another look!

src/sync/arc.rs
hawkw: yeah, i think doing it similarly to the way it's implemented in `std` is fine. thanks for the comment!

src/sync/arc.rs
```diff
@@ -25,8 +25,33 @@ impl<T> Arc<T> {
...
+            return Err(this);
+        }
+
+        assert_eq!(1, std::sync::Arc::strong_count(&this.value));
+        // work around our inability to destruct the object normally,
+        // because of the `Drop` presense.
+        this.obj.ref_dec(location!());
+        this.unregister();
+
+        // Use the same patter of unwrapping as `std` does.
```
hawkw, line 40: typo:
```suggestion
        // Use the same pattern of unwrapping as `std` does.
```

hawkw: also, it could be worth adding a github link to the similar code in std? not a blocker though.

tests/arc.rs
```diff
@@ -77,3 +77,13 @@ fn detect_mem_leak() {
         std::mem::forget(num);
     });
 }
+
+#[test]
+fn try_unwrap() {
```
hawkw, line 82: let's also add a test that `try_unwrap` fails when there's more than one ref?

hawkw (approved):
i made some minor additions (tests for the error case), and i'm going to go ahead and merge this PR when CI passes

kvark:
Thank you for adding tests and merging!

PR #286, opened 2022-07-10
Upgrade tracing subscriber and check minimum dependency versions in CI

faern:
Increase the minimum bound on the `tracing-subscriber` dependency in order to make `cargo +nightly update -Z minimal-versions && cargo check` work on this crate. See https://github.com/tokio-rs/tracing/pull/1890

I also added to CI to have it check the minimal versions. Do you like it this way? It's a bunch of duplication. But I felt it made sense to check the minimal versions with all the same tasks as the normal stable test.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,6 +43,25 @@ jobs:
       - name: Check
         run: cargo check --all-features
 
+  # Minimal dependency versions
+  minimal-dependency-versions:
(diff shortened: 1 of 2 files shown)


faern:
I did some extra pushes to make sure the CI job actually worked. It seems to do that. So now I have pushed the version upgrade again.

hawkw (approved):

Issue #290, opened 2022-10-17
Internal Panic: Index Out of Bounds


jswrenn:
The following program:
```rust
use loom::cell::Cell;

thread_local! {
    static ACTIVE_FRAME: Cell<()> = Cell::new(());
}

fn main() {
    loom::model(|| {
        let handle_a = loom::thread::spawn(|| {
            let _ = ACTIVE_FRAME.with(Cell::get);
        });
        let handle_b = loom::thread::spawn(|| ());
        handle_a.join().unwrap();
        handle_b.join().unwrap();
    });
}
```
...panics with:
```text
thread 'main' panicked at 'index out of bounds: the len is 2 but the index is 2', /Users/jswrenn/.cargo/registry/src/github.com-1ecc6299db9ec823/loom-0.5.6/src/rt/object.rs:286:25
stack backtrace:
   0: rust_begin_unwind
             at /rustc/ce7f0f1aa0f02c45cad0749e63f3086234b1f422/library/std/src/panicking.rs:556:5
   1: core::panicking::panic_fmt
             at /rustc/ce7f0f1aa0f02c45cad0749e63f3086234b1f422/library/core/src/panicking.rs:142:14
   2: core::panicking::panic_bounds_check
             at /rustc/ce7f0f1aa0f02c45cad0749e63f3086234b1f422/library/core/src/panicking.rs:84:5
   3: <usize as core::slice::index::SliceIndex<[T]>>::index_mut
             at /rustc/ce7f0f1aa0f02c45cad0749e63f3086234b1f422/library/core/src/slice/index.rs:259:14
   4: core::slice::index::<impl core::ops::index::IndexMut<I> for [T]>::index_mut
             at /rustc/ce7f0f1aa0f02c45cad0749e63f3086234b1f422/library/alloc/src/vec/mod.rs:2640:9
   5: <alloc::vec::Vec<T,A> as core::ops::index::IndexMut<I>>::index_mut
             at /rustc/ce7f0f1aa0f02c45cad0749e63f3086234b1f422/library/alloc/src/vec/mod.rs:2640:9
   6: loom::rt::object::Ref<T>::get_mut
             at /Users/jswrenn/.cargo/registry/src/github.com-1ecc6299db9ec823/loom-0.5.6/src/rt/object.rs:286:25
   7: loom::rt::cell::Cell::start_read::{{closure}}
[...]

PR #292, opened 2022-11-18
Track which reference belongs to with store

alexkeizer:
Add a unique id to stores, and keep track of this id in references.
When a reference is used with a different store, this will panic with a message explaining that the user might be trying to reuse object across executions.

This implements the suggestion in #290, although it adds a new id to stores, rather than reusing the execution ids.

One point I would like some more feedback on, is the naming of `Ref::from_usize`.
I changed the signature to also take a store id, but left the name as is. 
Changing it to `Ref::new` feels natural, but figured I'd ask for the rationale behind naming it `from_usize` in the first place.


alexkeizer:
I fixed the formatting, and the compile error for the checkpoint feature, but the `minrust` job complains that it can't build once_cell, and I'm not really sure how that's caused by my changes (I didn't touch dependencies).
The error is (taken from https://github.com/tokio-rs/loom/actions/runs/3498939232/jobs/5946011488)
```
Checking once_cell v1.16.0
error: edition 2021 is unstable and only available with -Z unstable-options.

error: could not compile `once_cell`
```

alexkeizer:
Some digging later, it seems that `once_cell` bumped its MSRV to 1.56.0 when going from 1.14 to 1.15, and loom transitively depends on `once_cell` 1.16, but `minrust` checks with rust 1.51.0.

Is there a specific procedure for changing looms MSRV, or should I just add a PR that changes the minrust job to check with 1.56?
Alternatively, do we want to stick with 1.51.0, and pin the once_cell dependency to 1.14 somehow?

PR #296, opened 2022-12-06
Bump MSRV to 1.59

alexkeizer:
Since dependencies changed their MSRV in a minor update, loom currently depends on `scoped-tls 1.0.1`, which has MSRV 1.59, making the minrust job fail.


hawkw (approved):

taiki-e:
Thanks!

PR #297, opened 2022-12-07
Fix drop order of tls and statics and avoid double panics on deadlock

Pointerbender:
Hello! I ran into some of the same issues as mentioned in #152, #179 and #291 and developed a fix for all of those issues. In a nutshell, the changes are:

- clarify the documentation of the mock `lazy_static` implementation regarding its `Drop` implementation
- ensure that global statics are dropped after thread locals
- ensure that `loom` does not panic inside a panic when dropping thread locals on the main thread
- add two regression tests for #152
- ensure that thread locals are destructed when a spawned thread panics
- ensure that loom doesn't double panic on deadlock
- add a test to verify that thread locals are properly destructed on spawned threads
- sprinkle comments generously everywhere to explain the code changes

This is my first time contributing to `loom`, please let me know if I can somehow improve this PR or if additional information is needed. Thank you!

PR #304, opened 2023-03-10
Avoid double panic when queued_spawn is dropped

adrianheine:
This is in the spirit of #297 on the quest to avoid double panics.


adrianheine:
The CI failure is rust-lang/rust#108988.

PR #310, opened 2023-05-18
Make thread stack size configurable

### LINKED ISSUE #309: Stack size of spawned threads is not configurable
The stack size of spawned threads is determined by the call to [`generator::Gn::new`](https://github.com/Xudong-Huang/generator-rs/blob/6e5cbeac4b3de27a93d8fa520db3200373cf27eb/src/gen_impl.rs#L230-L235) in `spawn_threads`: https://github.com/tokio-rs/loom/blob/16e5e9a0e562cf754ced04ac1a82802b8e492178/src/rt/scheduler.rs#L142-L158. Since loom doesn't specify a stack size, the `generator` crate's default is used, which is ~1000~ 4096 bytes. This isn't enough for code with deeply nested function calls; the result there is a stack overflow.

The stack size should be configurable either via environment variables or as an argument when invoking `loom::model`.

akonradi:
The default value of 4096 bytes is too small for some code. Make the value configurable via the Builder type, with the default value read from the environment variable LOOM_STACK_SIZE, or 4096 if it is not set.

Fixes #309

### DIFF
--- a/src/model.rs
+++ b/src/model.rs
@@ -1,6 +1,6 @@
 //! Model concurrent programs.
 
-use crate::rt::{self, Execution, Scheduler};
+use crate::rt::{self, Execution, Scheduler, ScheduleParams};
 use std::path::PathBuf;
 use std::sync::Arc;
 use std::time::{Duration, Instant};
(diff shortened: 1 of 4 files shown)


akonradi:
Closing this for now in favor of #311

PR #322, opened 2023-08-02
Add get_mut() to Mutex and RwLock

akonradi:
Add the standard library get_mut() method to Mutex and RwLock. This provides mutable access without locking for callers who already hold a &mut Mutex or &mut RwLock.

### DIFF
--- a/src/sync/mutex.rs
+++ b/src/sync/mutex.rs
@@ -58,6 +58,11 @@ impl<T> Mutex<T> {
         }
     }
 
+    /// Returns a mutable reference to the underlying data.
+    pub fn get_mut(&mut self) -> LockResult<&mut T> {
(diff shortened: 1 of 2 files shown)


src/sync/mutex.rs
```diff
@@ -58,6 +58,11 @@ impl<T> Mutex<T> {
         }
     }
 
+    /// Returns a mutable reference to the underlying data.
+    pub fn get_mut(&mut self) -> LockResult<&mut T> {
+        self.data.get_mut()
```
taiki-e, line 63:
```suggestion
        Ok(self.data.get_mut().unwrap())
```

taiki-e (commented):
Thanks!

taiki-e (approved):

PR #325, opened 2023-08-08
Implement AsRef and Borrow for Arc

notgull:
These traits are necessary to use the Arc in places that expect AsRef and Borrow.

### DIFF
--- a/src/sync/arc.rs
+++ b/src/sync/arc.rs
@@ -1,5 +1,6 @@
 use crate::rt;
 
+use std::borrow::Borrow;
 use std::pin::Pin;
 use std::{mem, ops, ptr};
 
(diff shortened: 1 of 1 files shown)


hawkw (approved):

src/sync/arc.rs
```diff
@@ -267,3 +268,17 @@ impl<T> From<T> for Arc<T> {
         Arc::new(t)
     }
 }
+
+impl<T: ?Sized> AsRef<T> for Arc<T> {
+    #[track_caller]
```
taiki-e, line 273:
```suggestion
```

src/sync/arc.rs
```diff
@@ -267,3 +268,17 @@ impl<T> From<T> for Arc<T> {
...
+    fn as_ref(&self) -> &T {
+        self
+    }
+}
+
+impl<T: ?Sized> Borrow<T> for Arc<T> {
+    #[track_caller]
```
taiki-e, line 280:
```suggestion
```

taiki-e (approved):
Thanks!

PR #339, opened 2024-03-16
Allow `Mutex` to work with `?Sized` types

Phantomical:
`std` supports this so it would be nice if loom did too. The only changes needed to do it is just adding `?Sized` where applicable.


Phantomical:
It looks like the `Option` in `RwLock` breaks this on 1.65 so I've changed it to only add `?Sized` for `Mutex`.

taiki-e (approved):
Thanks!

PR #341, opened 2024-03-28
feat: add loom::cell::Cell::into_inner

jbr:
Thanks for loom! I don't have sufficient context to know if there is a reason for this absence, but I ran into this difference between std::cell::Cell and loom::cell::Cell and it seemed easy enough to fix

### DIFF
--- a/src/cell/cell.rs
+++ b/src/cell/cell.rs
@@ -67,6 +67,12 @@ impl<T> Cell<T> {
     {
         self.replace(T::default())
     }
+
+    /// Unwraps the value, consuming the cell.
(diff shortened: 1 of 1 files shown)


taiki-e (approved):
Thanks.

notgull:
Any chance this change can be fit into a `loom` release?

taiki-e:
Already released in 0.7.2.
https://github.com/tokio-rs/loom/blob/master/CHANGELOG.md#072
https://docs.rs/loom/latest/loom/cell/struct.Cell.html#method.into_inner

PR #356, opened 2024-05-17
Implement `wait_while` for condvar

LucasSte:
**Problem**
I am using loom for concurrent tests, and `wait_while` was a feature I needed for the crate to be compatible with the project.

**Changes**
I implemented the `wait_while` function and added a test for it.


LucasSte:
I just found out that #167 also implements this. I'm happy to close this PR if that one is merged.

PR #358, opened 2024-05-20
Document what effect hints have on loom

udoprog:
This adds information which is otherwise buried in issues such as #162 to the documentation of the relevant methods.


src/rt/mod.rs
```diff
@@ -149,6 +149,89 @@ where
...
+///         let v2 = v1.clone();
+///
+///         let t1 = thread::spawn(move || {
+///             v2.lock();
+///             // critical section.
+///             v2.unlock();
+///         });
+///
+///         v1.lock();
+///         v1.unlock();
```
mox692, line 231: Perhaps we want to join the thread at the end?
```suggestion
///         v1.unlock();
///
///         t1.join().unwrap();
```

udoprog: I'll add it, but I also don't think it makes much of a difference.

The example code isn't testing any invariants, like correctly accessing the contents of an `UnsafeCell` and with the use of `loom::model` it should already execute all possible variants - including ones where the threads exits at the end. And here it doesn't matter since we don't have any code which depends on it.

mox692: Perhaps I need to read a little more loom code to be sure, but yeah I also think it wouldn't make much difference here.

taiki-e (approved):

PR #364, opened 2024-09-15
Bump MSRV to 1.73

taiki-e:
```
  error: package `generator v0.8.3` cannot be built because it requires rustc 1.77 or newer, while the currently active rustc version is 1.65.0
```
https://github.com/tokio-rs/loom/actions/runs/10867411903/job/30156136540#step:4:60

Depend on https://github.com/Xudong-Huang/generator-rs/pull/68

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ repository = "https://github.com/tokio-rs/loom"
 readme = "README.md"
 keywords = ["atomic", "lock-free"]
 categories = ["concurrency", "data-structures"]
-rust-version = "1.77"
+rust-version = "1.73"
 
 [features]
 default = []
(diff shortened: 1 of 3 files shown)


taiki-e:
```
  error: the `cargo::` syntax for build script output instructions was added in Rust 1.77.0, but the minimum supported Rust version of `generator v0.8.3` is 1.77.
  See https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script for more information about build script outputs.
  warning: build failed, waiting for other jobs to finish...
error: process didn't exit successfully: `rustup run 1.77 cargo check --all-features --manifest-path Cargo.toml` (exit status: 101)
```

Hmm...?

Darksonn (approved):
LGTM with path dep fixed

PR #370, opened 2024-12-27
Fix build error on not 64 bit arch

ryota2357:
`AtomicU64` and `AtomicI64` are defined only if `target_pointer_width = 64`

https://github.com/tokio-rs/loom/blob/3ef62e2011cb3b3baf76251f42b8c6c1a12d555d/src/sync/atomic/int.rs#L193-L197

But they are exported for all targets

https://github.com/tokio-rs/loom/blob/3ef62e2011cb3b3baf76251f42b8c6c1a12d555d/src/sync/atomic/mod.rs#L10-L12

### DIFF
--- a/src/sync/atomic/int.rs
+++ b/src/sync/atomic/int.rs
@@ -190,8 +190,8 @@ atomic_int!(AtomicI16, i16);
 atomic_int!(AtomicI32, i32);
 atomic_int!(AtomicIsize, isize);
 
-#[cfg(target_pointer_width = "64")]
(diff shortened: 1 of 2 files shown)


taiki-e:
Thanks for the PR. I think we should use `target_has_atomic = "64"` instead of `target_pointer_width = "64"`.

ryota2357:
Thanks, I changed to `target_has_atomic = 64`.

taiki-e (approved):

PR #373, opened 2025-01-26
Document Notify limitation

SpecificProtagonist:
`Notify` is limited to one thread at a time, but this isn't documented.

(I'm not sure why this limitation exists, but I'd be curious to know)


mox692 (approved):
Documenting current limitation sounds good to me. (I don't know the reason, though.)

PR #374, opened 2025-01-31
internal: support const on `UnsafeCell`

davidbarsky:
This method was `const` since 1.32 (https://doc.rust-lang.org/stable/std/cell/struct.UnsafeCell.html#method.new) and some code in [Salsa](https://github.com/salsa-rs/salsa/) makes use of creating `UnsafeCell`s in const contexts.

(Salsa is heavily used used by rust-analyzer and I'm spending a bit of time to make sure rust-analyzer's dependencies are reliable.)


davidbarsky:
I'll close this PR because I didn't realize it wasn't possible to const-ify some code here due to Loom's tracking code not being usable in const contexts.

PR #379, opened 2025-03-23
add `as_mut` to mutably access owned or mutably borrowed `UnsafeCell` safely

datdenkikniet:
See title.

This is the equivalent to [`core::cell::UnsafeCell::get_mut`][0], but with tracking. Instead of forcing unnecessary `unsafe` on users, we can simply give them mutable access to an owned value.

An example of where this causes unnecessary unsafe to be used can be found [here][1].

[0]: https://doc.rust-lang.org/std/cell/struct.UnsafeCell.html#method.get_mut
[1]: https://github.com/rtic-rs/rtic/pull/1038/files#r2009037113

### DIFF
--- a/src/cell/unsafe_cell.rs
+++ b/src/cell/unsafe_cell.rs
@@ -96,9 +96,24 @@ pub struct ConstPtr<T: ?Sized> {
 /// [here]: #correct-usage
 #[derive(Debug)]
 pub struct MutPtr<T: ?Sized> {
-    /// Drop guard representing the lifetime of the `ConstPtr`'s access.
+    /// Drop guard representing the lifetime of the `MutPtr`'s access.
     _guard: rt::cell::Writing,
     ptr: *mut T,
+}
+
+/// A checked mutable reference to an [`UnsafeCell`].
+///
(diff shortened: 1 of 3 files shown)


src/cell/unsafe_cell.rs
```diff
@@ -200,6 +215,21 @@ impl<T: ?Sized> UnsafeCell<T> {
...
+
+    /// Get a mutable reference to the wrapped value.
+    ///
+    /// This function returns a [`MutRef`] guard, which is analogous to a
+    /// `&mut T`, but tracked by Loom. As long as the returned `MutRef`
+    /// exists, Loom will consider the cell to be accessed mutably.
+    ///
+    /// This is analogous to [`core::cell::UnsafeCell::get_mut`].
+    #[track_caller]
+    pub fn as_mut(&mut self) -> MutRef<'_, T> {
```
datdenkikniet, line 227: I'm not entirely certain we even need tracking: if we can mutably borrow the value, we already have exclusive access to it, right? And because we borrow it mutably, no one can re-borrow it immutably until the borrow is dropped.

I've added the in-between struct so that someone can double-check my thoughts, but would be great to get some external input on it.

PR #391, opened 2025-09-21
Fix #254

### LINKED ISSUE #254: Invalid history with concurrent RMW and store?
Hello,

The failure of the below test puzzles me: since the RMW is atomic, I expect that the store can only be seen either before or after the RMW, so it should not be possible for a `swap` to read `false` and yet for the final state to also read `false`.

```rust

let flag = Arc::new(AtomicBool::new(false));

let th = thread::spawn({
    let flag = flag.clone();
    move || {
        // This unexpectedly triggers the final assertion.
        flag.swap(false, Ordering::Relaxed)

        // But this doesn't:
        //
        // flag.compare_exchange(true, false, Ordering::Relaxed, Ordering::Relaxed)
        //     .is_ok()
    }
});

flag.store(true, Ordering::Relaxed);

let pre_rmw_state = th.join().unwrap();
let final_state = flag.load(Ordering::Relaxed);

assert!(pre_rmw_state || final_state);
```

Interestingly, a simple CAS like in the above comment works as expected, but a CAS loop and other RMWs fail in a similar manner. Memory ordering does not appear to have any impact, nor atomic fences (not that I expected they would).

### LINKED ISSUE #389: Weird failing assertions
The following code fails the assertion and I am not sure, why.

Whenever the mutex is held by the second thread, either the first did its update (`a == b == false`) or it didn't (`a == b == true`).
The second thread may have run the compare-exchange (or the swap) after the first thread changed both values, but if either of them is successful, the assertion isn't run. If the function continues, both `a` and `b` must still have their old value.
Strangely, this logic appears to hold with the compare-exchange, but not with the swap (which should have the same semantics because these values are just booleans?).

```rust
#[cfg(test)]
#[cfg(feature = "loom")]
mod test {
    use std::sync::Arc;
    use loom::{
        thread,
        sync::{
            Mutex,
            atomic::{AtomicBool, Ordering}
        }
    };

    #[test]
    fn failing_test() {
        pub struct Foo
        {
            a: Mutex<bool>,
            b: AtomicBool,
        }

        loom::model(|| {
            let foo = Arc::new(Foo{
                a: Mutex::new(true),
                b: AtomicBool::new(true),
            });

            let thread_1 = {
                let foo = Arc::clone(&foo);
                thread::spawn(move || {
                    let maybe_guard = (*foo).a.lock();
                    let mut a = maybe_guard.unwrap();
                    foo.b.store(false, Ordering::Relaxed); // Was Release before, but that should not matter.
                    *a = false;
                })
            };

            let thread_2 = {
                let foo = Arc::clone(&foo);
                thread::spawn(move || {
                    // Either those don't have an effect (so the values remain equal), 
                    // or I am exiting the function (not running the assertion):
                    {
                        // works:
                        // if foo.b.compare_exchange(false, true, Ordering::Relaxed, Ordering::Relaxed).is_ok() { return; }

                        // causes assertion:
                        if !foo.b.swap(true, Ordering::Relaxed) { return; }
                    }

                    {
                        let maybe_guard = (*foo).a.lock();
                        let a = maybe_guard.unwrap();
                        let b = (*foo).b.load(Ordering::Relaxed);
                        assert!(*a == b);
                    }
                })
            };

            thread_1.join().unwrap();
            thread_2.join().unwrap();
        })
    }
}
```

Edit: This also happens when we replace the mutex by a typical spinlock and only do try_lock instead of lock. It appears as though the AtomicBool::swap is not actually atomic, but does separate read and write operations, and thereby loses the value written by the first thread.

ZobyTwo:
The issue was that we did not apply the RMW Atomicity constraint to the store, so it squeezed in between the read and the write part of the RMW.

Now, as soon as a RMW has happened (and observed the current value), all future stores must come after that in the modification order of that variable.

Fixes #254
Fixes #389


sbarral:
This also fixes https://github.com/tokio-rs/loom/issues/282, though only the [original test](https://github.com/tokio-rs/loom/issues/282#issue-1266791709) and not [the minimal repro](https://github.com/tokio-rs/loom/issues/282#issuecomment-2641386400) I then suggested (in other words, my minimal repro of #282 isn't one and triggers a different bug, but that's another story...).

Many thanks @ZobyTwo for your work on this, this fixes a bunch of other issues on my side. I hope someone more competent than me with the Loom internals can review this and get this merged :-)

PR #405, opened 2026-02-24
Add rust 1.91 AtomicPtr operations

RCoder01:
Adds `fetch_ptr_add`, `fetch_ptr_sub`, `fetch_byte_add`, `fetch_byte_sub`, `fetch_and`, `fetch_or`, and `fetch_xor` methods to `AtomicPtr`, all of which were stabilized in rust 1.91.


RCoder01:
Nvm I didn't see the MSRV. Maybe throw this on a pile of future work? Or maybe keep the methods but change their implementations to work with older rust versions?

PR #410, opened 2026-05-28
Fix #406

### LINKED ISSUE #406: Potential soundness issue in loom::lazy_static::Lazy::get returning static reference beyond execution lifetime
I found what appears to be a potential soundness issue in loom 0.7.2 related to loom::lazy_static::Lazy::get.

## Summary

`Lazy::get` returns `&'static T`, but the referenced value appears to be tied to loom runtime execution state and may be dropped when `loom::model` finishes. Safe code can move that `'static` reference outside the model closure and use it after execution teardown, which looks like a dangling-reference / use-after-free path.

### Why this looks problematic
* The API type promises a `'static` reference.
* Internally, the value lifetime is execution-scoped.
* The implementation uses lifetime transmutation while retrieving the value.
* Safe user code can store and use the returned reference after `loom::model` returns.

Minimal PoC:
```rust
use loom::lazy_static::Lazy;
use std::marker::PhantomData;
use std::sync::{Arc, Mutex};

static MALICIOUS_LAZY: Lazy<String> = Lazy {
    init: || String::from("Exploit UAF"),
    _p: PhantomData,
};

fn main() {
    let leaked_ref = Arc::new(Mutex::new(None));
    let leaked_clone = leaked_ref.clone();

    loom::model(move || {
        let s: &'static String = MALICIOUS_LAZY.get();
        *leaked_clone.lock().unwrap() = Some(s);
    });

    let uaf_string: &'static String = leaked_ref.lock().unwrap().unwrap();
    println!("UAF Read: {}", uaf_string);
}
```

### Observed behavior with cargo run

The output of `cargo run` is different when trying some times:

Image

So is this indicates an unsound problem in this crate? Thank you for your reply!

yilin0518:
This pull request adds safety documentation to the mock `lazy_static` implementation, clarifying the lifetime and usage constraints of references returned during loom model execution. This PR aims to solve the #406 by adding "#safety" section to remind users the caveat.

Documentation improvements:

* Added a detailed safety section to the `Lazy::get` method, explaining that the returned `&'static T` reference is only valid for the duration of the current loom execution and must not escape it.
* Updated the crate-level documentation in `src/lib.rs` to warn users that references from loom's mock `lazy_static` are only valid during a modeled execution.

### DIFF
--- a/src/lazy_static.rs
+++ b/src/lazy_static.rs
@@ -26,6 +26,13 @@ pub struct Lazy<T> {
 
 impl<T: 'static> Lazy<T> {
     /// Mock implementation of `lazy_static::Lazy::get`.
+    ///
+    /// # Safety
+    /// The returned reference is typed as `&'static T`, but the underlying value
+    /// only lives for the duration of the current loom execution. Callers must
+    /// ensure that the reference does not escape the current execution, such as
+    /// by storing it in state that is later accessed after [`crate::model()`]
+    /// returns.
     pub fn get(&'static self) -> &'static T {
         // This is not great. Specifically, we're returning a 'static reference to a value that
         // only lives for the duration of the execution. Unfortunately, the semantics of lazy
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -109,6 +109,10 @@
 //! tests, the `std` should be used, since the loom runtime won't be active. This means that
 //! library code will need to use conditional compilation to decide which types to use.
 //!
+//! Loom's mock `lazy_static` support is only valid while a modeled execution is
+//! active. References obtained from it must not escape the current
+//! [`model`](mod@model) execution.
+//!
 //! It is recommended to use a `loom` cfg flag to signal using the loom types. You can do this by
 //! passing `RUSTFLAGS="--cfg loom"` as part of the command when you want to run the loom tests.
 //! Then modify your `Cargo.toml` to include loom like this:


src/lazy_static.rs
```diff
@@ -26,6 +26,13 @@ pub struct Lazy<T> {
...
 impl<T: 'static> Lazy<T> {
     /// Mock implementation of `lazy_static::Lazy::get`.
+    ///
+    /// # Safety
+    /// The returned reference is typed as `&'static T`, but the underlying value
+    /// only lives for the duration of the current loom execution. Callers must
+    /// ensure that the reference does not escape the current execution, such as
+    /// by storing it in state that is later accessed after [`crate::model()`]
+    /// returns.
     pub fn get(&'static self) -> &'static T {
```
Darksonn, line 36: It doesn't make sense to place a safety section on a safe method.

src/lib.rs
```diff
@@ -109,6 +109,10 @@
 //! tests, the `std` should be used, since the loom runtime won't be active. This means that
 //! library code will need to use conditional compilation to decide which types to use.
 //!
+//! Loom's mock `lazy_static` support is only valid while a modeled execution is
+//! active. References obtained from it must not escape the current
+//! [`model`](mod@model) execution.
```
Darksonn, line 114: This is a more general concern. No loom types should be reused from one model execution to another.

src/lazy_static.rs
yilin0518: > It doesn't make sense to place a safety section on a safe method.

Is it appropriate for marking this method as unsafe? Because this method can cause panic with improper usage.

PR #419, opened 2026-07-22
Rename `fetch_update` to `try_update`

johamster:
This is in line with `std` renaming the functions and deprecating the old `fetch_update` API (which will reach stable with Rust 1.99).

### DIFF
--- a/src/sync/atomic/bool.rs
+++ b/src/sync/atomic/bool.rs
@@ -126,7 +126,7 @@ impl AtomicBool {
     }
 
     /// An alias for [`Self::try_update`].
-    #[deprecated = "renamed to try_update for consistency"]
+    #[deprecated = "renamed to `try_update` for consistency"]
     #[track_caller]
     pub fn fetch_update<F>(
         &self,
--- a/src/sync/atomic/bool.rs
+++ b/src/sync/atomic/bool.rs
@@ -113,6 +113,21 @@ impl AtomicBool {
     /// a [`Result`] of [`Ok`]`(previous_value)` if the function returned [`Some`]`(_)`, else
     /// [`Err`]`(previous_value)`.
     #[track_caller]
+    pub fn try_update<F>(
+        &self,
+        set_order: Ordering,
+        fetch_order: Ordering,
+        f: F,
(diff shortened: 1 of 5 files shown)


src/sync/atomic/bool.rs
```diff
@@ -113,6 +113,21 @@ impl AtomicBool {
     /// a [`Result`] of [`Ok`]`(previous_value)` if the function returned [`Some`]`(_)`, else
     /// [`Err`]`(previous_value)`.
     #[track_caller]
+    pub fn try_update<F>(
```
johamster, line 116: FYI: I simply renamed the existing `fetch_update` functions and then added a shallow wrapper with the old `fetch_update` name below to not break existing code. The old name is marked `deprecated`.

src/sync/atomic/bool.rs
```diff
@@ -113,6 +113,21 @@ impl AtomicBool {
...
+        f: F,
+    ) -> Result<bool, bool>
+    where
+        F: FnMut(bool) -> Option<bool>,
+    {
+        self.0.try_update(set_order, fetch_order, f)
+    }
+
+    /// An alias for [`Self::try_update`].
+    #[deprecated = "renamed to try_update for consistency"]
```
wvwwvwwv: #[deprecated(
      since = "1.99.0",
      note = "renamed to `try_update` for consistency",
      suggestion = "try_update"
  )]
-> https://doc.rust-lang.org/src/core/sync/atomic.rs.html#1996

johamster: Right, I can add the "suggestion" but would not add the version number. `loom` has different versioning. Not sure it is really important to specify the version.

johamster: Wait, `suggestion` is unstable (see [`deprecated_suggestion`](https://github.com/rust-lang/rust/issues/94785)). I'd leave this to the maintainer to decide whether we want to activate the unstable feature.

wvwwvwwv: Then, please also consider removing this warning, as many users may continue using older versions of Rust.

